PR #1, opened 2024-02-17
Make the /nl GUI easier to use

okx-code:
- Sort all groups by their rank, and then their name
- Sort members of groups by their rank, and then their name
- Show /nlpp and /nlrm command hints (most people don't seem to be aware of these commands)


awoo-civ:
1. Any reason the group sorting isn't also sorted by rank first?
2. Would be nice to have this change apply to `/nllm` too.

PR #2, opened 2024-02-17
Refactor DogFacts to use MiniMessage

Protonull:
Moving https://github.com/CivMC/SimpleAdminHacks/pull/76 over to here.


okx-code (approved):
looks good, but does dogfacts really broadcast at random intervals, determined on restart? wtf?

RedDevel2:
Really? Then why is there an interval time in the config?

okx-code:
Oh it's just the inital delay, not the interval. Still weird but ok I guess

Protonull:
Is there anything in particular preventing this from being merged since it's missed two assigned milestones?

okx-code:
The current dog facts config would need to be changed

Protonull:
> The current dog facts config would need to be changed

Done

okx-code (approved):

PR #235, opened 2024-02-19
Welcome book command

MrJeremyFisher:
Moves the previous `/introbook` command to `/giveintrobook` and makes `/introbook` display a copy of the player's welcome book.


RedDevel2:
Very nice 👍

PR #237, opened 2024-02-20
Add Map Copy Protection

MrJeremyFisher:
Maps can be copy protected by running `/copyprotect` while holding a filled map. Copy protected maps may still be locked and have their scale changed, but may not be copied in either a cartography table or a crafting interface.

Same as https://github.com/CivMC/SimpleAdminHacks/pull/80. Moved over because of the switch to a monorepo.


AngrySoundTech (commented):
👍

PR #238, opened 2024-02-20
Prevent Innapropriate Flower Pot Modification

MrJeremyFisher:
Block removing and placing plants in flower pots if the player doing so does not have the appropriate permissions


AngrySoundTech (commented):
👍

PR #257, opened 2024-02-20
Add Kotlin support

Protonull:
Standardises any Kotlin support

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,10 +1,12 @@
 import com.gradle.enterprise.gradleplugin.GradleEnterpriseExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinTopLevelExtension
 import xyz.jpenilla.runpaper.task.RunServer
 
 plugins {
     id("io.papermc.paperweight.userdev") version "1.5.10" apply false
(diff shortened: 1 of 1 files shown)


okx-code:
What is the use case for this?

Protonull:
> What is the use case for this?

Being able to code in Kotlin?

okx-code:
What specifically do you want to use this for?

AngrySoundTech:
This should be added with something that uses it, not on it's own

Protonull:
> What specifically do you want to use this for?

1. For when Kira gets added.

2. While I know there was a discussion about not mixing Java and Kotlin within the same plugin, I nonetheless think converting the `NBTTagCompoundExtensions` class (and its ilk) over to Kotlin's extension methods is the only sensible option. Very little will change from Java's perspective, but it'll be all the better for any plugin that has Kotlin code and needs to use those utilities.

3. I wanna refactor ItemExchange into Kotlin :) I know this is kind of a meme, but if we can't mix Java and Kotlin within the same plugin, and refactoring a plugin into Kotlin is a meme, then is Kotlin support only being added for Kira and that's it?

Issue #253, opened 2024-02-20
/nltg doesn't promote a player to owner


okx-code:
If someone is a primary owner but not an owner they lock themselves out of a group

PR #262, opened 2024-02-21
Fix primary owner transfer not being set to OWNER role

Diet-Cola:


plugins/namelayer-paper/src/main/java/vg/civcraft/mc/namelayer/command/commands/TransferGroup.java
```diff
@@ -66,9 +66,11 @@ public static boolean attemptTransfer(Group g, Player owner, UUID futureOwner) {
 					+ " is not a member of the group and can't be made primary owner!");
 			return false;
 		}
+		g.removeMember(futureOwner);
+		g.addMember(futureOwner, GroupManager.PlayerType.OWNER);
```
Protonull, line 70: Just thought I'd put this here :P 
image

Diet-Cola: Best part about this is, I followed the logic of the promote player command... which doesnt even use this either

Protonull: Civ code D:

Diet-Cola:
Fixes #253

PR #275, opened 2024-02-21
Optimise snitch GUI

okx-code:
- Only render the snitches for the current page
- Use components instead of adding text lore (this is about 90% faster than the previous method)
- Also fix the plugin name (previously it was jukealert-paper, now it correctly shows as JukeAlert)

### DIFF
--- a/plugins/jukealert-paper/src/main/java/com/untamedears/jukealert/gui/SnitchOverviewGUI.java
+++ b/plugins/jukealert-paper/src/main/java/com/untamedears/jukealert/gui/SnitchOverviewGUI.java
@@ -2,12 +2,21 @@ package com.untamedears.jukealert.gui;
 
 import com.untamedears.jukealert.model.Snitch;
 import com.untamedears.jukealert.model.appender.DormantCullingAppender;
+
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.format.TextColor;
+import net.kyori.adventure.text.format.TextDecoration;
 import net.md_5.bungee.api.ChatColor;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.meta.ItemMeta;
+import vg.civcraft.mc.civmodcore.chat.ChatUtils;
 import vg.civcraft.mc.civmodcore.inventory.gui.Clickable;
+import vg.civcraft.mc.civmodcore.inventory.gui.FastMultiPageView;
 import vg.civcraft.mc.civmodcore.inventory.gui.IClickable;
 import vg.civcraft.mc.civmodcore.inventory.gui.MultiPageView;
 import vg.civcraft.mc.civmodcore.inventory.items.ItemUtils;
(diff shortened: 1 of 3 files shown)


RedDevel2:
Now that I think about it. The name is probably why ItemExchange didn't recognize it. 💀

okx-code:
Yep it broke for me as well

plugins/jukealert-paper/src/main/java/com/untamedears/jukealert/gui/SnitchOverviewGUI.java
```diff
@@ -29,38 +38,39 @@ public SnitchOverviewGUI(Player player, List<Snitch> snitches, String title, boo
...
-				MetaUtils.addLore(meta, ChatColor.YELLOW + "Group: " + snitch.getGroup().getName());
+				List<Component> lore = new ArrayList<>();
+				lore.add(ChatUtils.newComponent("Location: ").color(NamedTextColor.AQUA)
+						.append(Component.text(location.getWorld().getName() + " ").color(NamedTextColor.WHITE))
+						.append(Component.text(location.getBlockX()).color(NamedTextColor.RED))
+						.append(Component.text(", ").color(NamedTextColor.WHITE))
+						.append(Component.text(location.getBlockY()).color(NamedTextColor.GREEN))
+						.append(Component.text(", ").color(NamedTextColor.WHITE))
+						.append(Component.text(location.getBlockZ()).color(NamedTextColor.BLUE)));
+				lore.add(ChatUtils.newComponent("Group: " + snitch.getGroup().getName()).color(NamedTextColor.YELLOW));
```
Protonull, line 58: You should do the same with all of these components too :]

okx-code: Nah

Protonull:
> Now that I think about it. The name is probably why ItemExchange didn't recognize it. 💀

That could be the cause of some unintended plugin-load ordering, but it should NOT cause the error. JukeAlert is supposed to be a soft-dependency; ItemExchange is suppose to work even if JukeAlert isn't there.

The error is also completely baffling, it's a `NoClassDefFoundError`, which means a class is being used in a non-reflective manner (eg: an instance is being constructed with `new`), but the class isn't present.

The missing class is `LoggableAction`, right?. That by itself is bizarre because nothing in ItemExchange references `LoggableAction`: the class `ShopPurchaseAction` extends `LoggablePlayerAction` and *that* implements `LoggableAction`, so... why is the missing class `LoggableAction` and not `LoggablePlayerAction`? Those are both JukeAlert classes.

Also, `ShopPurchaseAction` is only interacted with in two places:

1. Inside a lambda that's registered as an action-factory to JukeAlert, something that only happens if JukeAlert is enabled.

2. Inside an event listener that's only registered if JukeAlert is enabled.

So how in the hell is this class being used?!

Protonull:
Ohhhh, okay, just figured out what's happening: the action-factory lambda gets hoisted by javac into a sibling method:
image

So when the class is evaluated, it finds the method and tries to evaluate its return type, LoggableAction, which isn't loaded. The only fix to this is to deliberately avoid the lambda-isation and turn it into a sibling class.

PR #311, opened 2024-02-22
Exilepearl 1.20

MrJeremyFisher:


plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/core/ExilePearlCore.java
```diff
@@ -562,6 +569,11 @@ public String getName() {
 		return plugin.getName();
 	}
 
+	@Override
+	public @NotNull LifecycleEventManager<Plugin> getLifecycleManager() {
+		throw new UnsupportedOperationException();
```
okx-code, line 574: plugin.getLifecycleManager?

PR #317, opened 2024-02-22
Add NameColors

okx-code:


plugins/namecolors-paper/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id("io.papermc.paperweight.userdev")
+}
+
+version = "2.0.0-SNAPSHOT"
+
+repositories {
+    maven("https://repo.kryptonmc.org/releases")
+}
```
AngrySoundTech, line 9: We should move all the repositories to the main file instead of spreading them out

AngrySoundTech (changes requested):

AngrySoundTech (approved):

PR #319, opened 2024-02-23
Add a toggle to enable/disable banned pearls in ep showall

walkersgaming:
Added a button to `/ep showall` menu to toggle banned pearls and turned them off by default. Toggle shares cooldown with command as it essentially reruns the command each time it's clicked but with the filter on or off.

https://github.com/CivMC/Civ/assets/14056973/4b3cd5af-6b47-428a-b658-6775cddc3fe7

### DIFF
--- a/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
+++ b/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
@@ -18,6 +18,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.format.TextDecoration;
 import net.md_5.bungee.api.ChatColor;
 import org.apache.commons.collections4.ComparatorUtils;
 import org.apache.commons.collections4.list.LazyList;
@@ -43,6 +44,7 @@ public class CmdShowAllPearls extends PearlCommand {
 	private static final DateFormat DATE_FORMAT = new SimpleDateFormat("dd MMM yyyy");
 	private static final Map<UUID, Long> COOLDOWNS = new HashMap<>();
 	private static final long COOLDOWN = 10_000; // 10 seconds
+	private boolean bannedPearlToggle = true;
 
 	public CmdShowAllPearls(final ExilePearlApi pearlApi) {
 		super(pearlApi);
@@ -65,6 +67,10 @@ public class CmdShowAllPearls extends PearlCommand {
 		}
 		COOLDOWNS.put(sender.getUniqueId(), now);
 
+		generateOpenPearlsMenu(sender);
+	}
+
+	private void generateOpenPearlsMenu(Player sender) {
 		final Location senderLocation = sender.getLocation();
 		final double pearlExclusionRadius = this.plugin.getPearlConfig().getRulePearlRadius() * 1.2;
 		final boolean isBanStickEnabled = this.plugin.isBanStickEnabled();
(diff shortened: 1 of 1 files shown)


plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
```diff
@@ -43,6 +44,7 @@ public class CmdShowAllPearls extends PearlCommand {
 	private static final DateFormat DATE_FORMAT = new SimpleDateFormat("dd MMM yyyy");
 	private static final Map<UUID, Long> COOLDOWNS = new HashMap<>();
 	private static final long COOLDOWN = 10_000; // 10 seconds
+	private boolean bannedPearlToggle = true;
```
Protonull, line 47: Just keep in mind this is a *global* toggle, so if someone hides banned players in their `/ep showall` window, then everyone else after that will also have banned players hidden. And when someone then unhides them, everyone after that will also see them.

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
```diff
@@ -188,7 +205,39 @@ public void addedToInventory(ClickableInventory inv, int slot) {
...
 
 		LazyList<IClickable> lazyContents = MoreCollectionUtils.lazyList(contentSuppliers);
-		new MultiPageView(sender, lazyContents, "All Pearls", true).showScreen();
+		final var pageView = new MultiPageView(sender, lazyContents, "All Pearls", true);
+
+		pageView.setMenuSlot(contructBannedPearlsToggleClick(), 4);
+		pageView.showScreen();
 	}
 
+	private IClickable contructBannedPearlsToggleClick() {
```
Protonull, line 214: contruct :]

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
walkersgaming: Huh, didn't realise. Is there a better way to do it then?

Protonull: You could do something similar to the `COOLDOWNS` map slightly above. Or you could make it a player setting and let people toggle it on or off without needing to run `/ep showall`, which might be preferred since it's a particularly taxing command to run. It'll also help with option persistence, as in, you don't need to open `/ep showall` and then re-open it with the toggle, doubling the strain on the server, to see what you want to see.

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
```diff
@@ -43,6 +44,7 @@ public class CmdShowAllPearls extends PearlCommand {
 	private static final DateFormat DATE_FORMAT = new SimpleDateFormat("dd MMM yyyy");
 	private static final Map<UUID, Long> COOLDOWNS = new HashMap<>();
 	private static final long COOLDOWN = 10_000; // 10 seconds
+	private static final Map<UUID, Boolean> TOGGLES = new HashMap<>();
```
okx-code, line 47: This could be a HashSet

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
```diff
@@ -188,7 +207,39 @@ public void addedToInventory(ClickableInventory inv, int slot) {
...
+							.decoration(TextDecoration.ITALIC, false)
+							.color(NamedTextColor.AQUA)
+							.content("Currently turned " + (!bannedPearlToggle ? "on" : "off"))
+							.build());
+			return true;
+		});
+		return new Clickable(item) {
+			@Override
+			public void clicked(final Player clicker) {
+				if (onCoolDown(clicker)) return;
```
okx-code, line 235: Does this mean that the toggle can only clicked 10 seconds after opening the GUI?

walkersgaming: Yes but it will only matter on the rare occasion someone wants to look at banned pearls

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/command/CmdShowAllPearls.java
walkersgaming: Updated with the former idea

okx-code (approved):

RedDevel2:
On testing it now never shows banned players in /ep showall
The button to switch, always says "Currently turned on" no matter if you switch it.
Likewise the chat message always says "Banned pearls toggled off" whenever you click the switch. 

Also the /ep showall command now itself has a cooldown? I doubt that's supposed to be a thing.

okx-code:
Should be fixed now

walkersgaming:
I'm not sure that should've been changed, as the variable hadn't been updated with the toggle change, hence it was inverted

PR #321, opened 2024-02-23
Replace MoreClassUtils.castOrNull() with instanceof pattern matching

Protonull:


plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/utilities/MoreClassUtils.java
okx-code, line 1: Why not delete this whole file?

Protonull: I intend to add some utilities to it soon in another PR since they're outside the scope of this PR, and I figured deleting the file then remaking it is probably a bit extra, but I can amend this PR to delete the file if that's what you really want.

Protonull: Okay, if you're sure

PR #329, opened 2024-02-24
Add workflows for PR checks

AngrySoundTech:
PR Comments will not work until they're on `main`, they should be good though since I copied them from our old projects


okx-code:
Can you explain what this does

AngrySoundTech:
> Can you explain what this does

This PR Enables checks automatically to run on all PRs, which will run `gradle check`, and `yamllint` (errors currently ignored until we fix them all).

You can see the checks ran on this PR as a requirement before merging, taking 1 minute to ensure that the project builds, tests pass, and yaml files are formatted correctly (which will be more important especially when plugin config gets moved here). If we add code style checks and such to gradle, that will also be part of the check task

It also adds a workflow that comments with an easy link to the built artifact on the pr for testing, the output of which can be seen here: https://github.com/CivMC/CivModCore/pull/98#issuecomment-1849098604. Once this is merged that will start working, since the workflow needs to be on `main`

okx-code (approved):

PR #335, opened 2024-02-27
Fix horse stats message double sending

MrJeremyFisher:
The `PlayerInteractEntityEvent` event fires twice, once for both of the players hands, so the message was displayed twice. Only occurs if the stats are checked whilst another player is riding the horse, but ignoring the offhand version of the event fixes it.


okx-code (approved):

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/HorseStats.java
```diff
@@ -32,6 +33,9 @@ public void onHorseStatCheck(PlayerInteractEntityEvent event) {
 		if (!item.getType().equals(config.getHorseCheckerItem())) {
 			return;
 		}
+		if (event.getHand() == EquipmentSlot.OFF_HAND) {
```
Protonull, line 36: Maybe I'm just paranoid, but I'd defensively replace this with `if (event.getHand() != EquipmentSlot.HAND)` instead.

PR #338, opened 2024-02-28
Deoxidise as many blocks as the tool in-hand can

xFier:
Changes the deoxidisation to process as many steps as it can before the tool breaks.


okx-code (changes requested):
Comparing damage values of an axe is not really a good way of checking durability, because the Unbreaking enchantment exists. This should be a while loop that applies damage and deoxidises one at a time, until the tool is broken or it's fully deoxidied.

RedDevel2 (approved):

PR #345, opened 2024-02-28
A partial fix for minecart derailment

okx-code:
If a minecart is going to derail because it is going too fast, this fix will backtrack its movement and move it as if it was going 8m/s. This will slow down the minecart slightly but prevent derailment.

This does not fix fast minecarts being unable going up hills or minecarts jumping across tracks.


AngrySoundTech (commented):
Code looks like it makes sense, does this hack need a config value added as well though?

okx-code:
Yep you're right

RedDevel2 (approved):

PR #351, opened 2024-03-02
Prevent grass spread to reinforced dirt

MrJeremyFisher:


AngrySoundTech (approved):
This looks good to me, I assume reinforcement caching would make this not too intensive? Grass seems to try to spread more than any other of our checks here that we prevent.

If someone else thinks that's sane too then we should be good.

MrJeremyFisher:
Yeah, I tried some testing on a large area of grass & dirt and didn't see too much performance impact. Unsure how it would perform on live though obviously

plugins/citadel-paper/src/main/java/vg/civcraft/mc/citadel/listener/BlockListener.java
```diff
@@ -198,6 +199,22 @@ public void onBlockFromToEvent(BlockFromToEvent event) {
 		}
 	}
 
+	// prevent "enemy" grass spreading to reinforced dirt
+	@EventHandler(priority = EventPriority.LOW, ignoreCancelled = true)
+	public void onGrassSpread(BlockSpreadEvent event) {
+		if (event.getSource().getType() != Material.GRASS_BLOCK) return;
+
+		Reinforcement destRein = ReinforcementLogic.getReinforcementProtecting(event.getBlock());
+
```
awoo-civ, line 208: Nit: Remove newline, the if statement pertains directly to the variable declared above, they're tightly related so they should be tightly visually placed as well.

plugins/citadel-paper/src/main/java/vg/civcraft/mc/citadel/listener/BlockListener.java
```diff
@@ -198,6 +199,22 @@ public void onBlockFromToEvent(BlockFromToEvent event) {
...
+		if (destRein != null) {
+			Reinforcement sourceRein = ReinforcementLogic.getReinforcementProtecting(event.getSource());
+
```
awoo-civ, line 211: Nit: Remove newline, same as above.

plugins/citadel-paper/src/main/java/vg/civcraft/mc/citadel/listener/BlockListener.java
```diff
@@ -198,6 +199,22 @@ public void onBlockFromToEvent(BlockFromToEvent event) {
...
 	}
 
+	// prevent "enemy" grass spreading to reinforced dirt
+	@EventHandler(priority = EventPriority.LOW, ignoreCancelled = true)
+	public void onGrassSpread(BlockSpreadEvent event) {
+		if (event.getSource().getType() != Material.GRASS_BLOCK) return;
+
+		Reinforcement destRein = ReinforcementLogic.getReinforcementProtecting(event.getBlock());
+
+		if (destRein != null) {
```
awoo-civ, line 209: Perhaps turn this into an early return instead of nesting, i.e. `if (destRein == null) { return }`.

okx-code (approved):

PR #356, opened 2024-03-04
Create PVP Server Container

AngrySoundTech:
This PR Introduces container building to the Monorepo, and adds tooling to support as such.

There are still some things that need to be worked out after this pr:

- [ ] Sharing resources/configs (ops.yml, etc) between multiple containers
- [x] Building just a container or just a plugin from PRs or tags
- [ ] Database schemas for multiple servers, for plugins that are not designed to share a database
- [x] And probably something else that I thought of before but forgot before I got this far.


README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Current services and exposed ports are:
+
+| Name     | Ports     | Profile    |
+|----------|-----------|------------|
+| pvp      |           |            |
+| mariadb  | TCP/3306  |            |
+| postgres | TCP/5432  |            |
+| rabbitmq | HTTP/5672 |            |
+| grafana  | HTTP/3000 | monitoring |
```
okx-code, line 24: should be all tcp

okx-code (approved):

README.md
AngrySoundTech: I wanted to specifically call out the service not the protocol to clue people into that being what it is, especially since people might not be familiar with the rabbitmq admin gui, but I can change this to be more clear

Protonull: Might be worth adding the RabbitMQ dashboard port too: [15672](https://www.rabbitmq.com/docs/management#usage-ui)

PR #371, opened 2024-03-15
Make sapling hack better

Diet-Cola:
Fix NPE when the config doesnt have a leaf block defined, also leave a comment to be found when updating so hopefully that gets caught and updated. Is there a way to do this better though?

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/SaplingHack.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/SaplingHack.java
@@ -5,6 +5,7 @@ import com.programmerdan.minecraft.simpleadminhacks.configs.SaplingConfig;
 import com.programmerdan.minecraft.simpleadminhacks.framework.SimpleHack;
 import java.util.Random;
 import org.bukkit.Material;
+import org.bukkit.Tag;
 import org.bukkit.block.Block;
 import org.bukkit.configuration.ConfigurationSection;
 import org.bukkit.enchantments.Enchantment;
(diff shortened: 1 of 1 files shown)


plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/SaplingHack.java
```diff
@@ -43,7 +44,7 @@ public void onBlockBreak(BlockBreakEvent event) {
 			return;
 		}
 		double nextDouble = random.nextDouble();
-		if (nextDouble <= this.config.getChanceMap().get(brokenBlock.getType())) {
+		if (nextDouble <= this.config.getChanceMap().getOrDefault(brokenBlock.getType(), 0.0D)) {
```
okx-code, line 47: An NPE here will never happen because of the previous check

okx-code (approved):

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/SaplingHack.java
```diff
@@ -56,17 +54,14 @@ public void onLeafDecay(LeavesDecayEvent event) {
...
 		}
 		double nextDouble = random.nextDouble();
-		if (nextDouble <= this.config.getChanceMap().get(decayedBlock.getType())) {
+		if (nextDouble <= this.config.getChanceMap().getOrDefault(decayedBlock.getType(), 0.0D)) {
 			decayedBlock.getWorld().dropItemNaturally(decayedBlock.getLocation(), new ItemStack(getSaplingMaterial(decayedBlock.getType()), 1));
 			logger.info("Dropped extra " + getSaplingMaterial(decayedBlock.getType())  + " from being broken at " + decayedBlock.getLocation());
 		}
 	}
 
 	private boolean checkIfLeafBlock(Block block) {
```
okx-code, line 63: Perhaps now this method can be removed?

PR #373, opened 2024-03-16
Prevent boats from entering minecarts

okx-code:
Fixes https://bugs.mojang.com/browse/MC-113871


AngrySoundTech (commented):
The code looks fine, but I'm not sure we should change this. It's on Mojang's bug tracker, but has been part of the game for the past seven years and I don't see any reason to disable it.

PR #374, opened 2024-03-16
Prevent respawn anchors being used as anti-place

okx-code:
This hack prevents you from right clicking respawn anchors at all, this fix makes sure that it only prevents the respawn anchor from exploding, but doesn't prevent you placing blocks.


AngrySoundTech (commented):
Doesn't this defeat the point of the hack, and let you explode anchors in the overworld?

okx-code:
You're right, exploding anchors needs to be disabled, but it doesn't defeat the point of the hack as it would still work in the nether.

AngrySoundTech (approved):

PR #376, opened 2024-03-16
Allow interacting with chests and trapdoors if doing so would not actually open them

okx-code:
Previously players were not able to use potions on top of these kinds of blocks or place blocks on top of them with shift right click. This change will allow that behaviour.


AngrySoundTech (commented):
I think a lot of players have built around this as expected functionality, and I'm not sure we should change it or not. As a feature I have no strong opinions either way, it just depends on the design philosophy of citadel and how we think that _any_ interactions with reinforced blocks should go. Should probably come to a consensus as a team on it

AngrySoundTech (approved):

PR #385, opened 2024-03-22
Replace HashSet with OpenObjectHashSet

okx-code:
A quick fix for the large memory usage of this object, HashSet uses an empty object as a value of each element in its set, but OpenObjectHashSet doesn't rely on HashMap, has no values, and is therefore more memory efficient.


AngrySoundTech (approved):
I'm not researching this myself, I'll just trust it makes it better.

PR #386, opened 2024-03-23
Fix portal hack not triggering

okx-code:
If you are falling really fast, you will be falling multiple blocks per second, and could possibly pass straight through an end portal. This fix changes the hack to use an event triggered earlier to detect if a player passes through an end portal.

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/PortalModifyHack.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/PortalModifyHack.java
@@ -1,5 +1,6 @@
 package com.programmerdan.minecraft.simpleadminhacks.hacks.basic;
 
+import com.destroystokyo.paper.event.server.ServerTickEndEvent;
 import com.programmerdan.minecraft.simpleadminhacks.SimpleAdminHacks;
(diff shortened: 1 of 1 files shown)


plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/PortalModifyHack.java
```diff
@@ -40,18 +48,27 @@ public void onPlayerEnterPortal(PlayerPortalEvent event) {
...
+	@EventHandler
+	public void onTick(ServerTickEndEvent event) {
+		for (Map.Entry<Player, Location> entry : teleports.entrySet()) {
+				entry.getKey().teleport(entry.getValue(), PlayerTeleportEvent.TeleportCause.END_PORTAL);
+				if (entry.getValue().getWorld().getName().equals(targetWorld)) {
+					spawnExit(entry.getValue());
+				}
 		}
+		teleports.clear();
 	}
```
AngrySoundTech, line 72: Is it ever possible that this event can be fired simultaneously, such as during low TPS? If so I could imagine it causing some problems, and potentially more players falling through portals than before.

okx-code: Simultaneously with what? It runs at the end of a tick.

RedDevel2 (approved):

PR #387, opened 2024-03-23
Add logging for ChunkMeta

okx-code:
Hopefully this can be used to diagnose the issue

### DIFF
--- a/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
+++ b/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
@@ -7,24 +7,26 @@ import co.aikar.commands.annotation.Description;
 import co.aikar.commands.annotation.Subcommand;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.config.Configurator;
 import org.bukkit.command.CommandSender;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 @CommandAlias("cmc")
 public class ChunkMetaCommand extends BaseCommand {
 
-	private static final Logger CHUNK_META_LOGGER = Logger.getLogger("Chunk meta");
+	private static final Logger CHUNK_META_LOGGER = LogManager.getLogger("Chunk meta");
 
 	@Subcommand("togglechunkmeta")
 	@Description("Toggle showing chunk meta logs")
 	@CommandPermission("cmc.debug")
 	public void chunkmeta(CommandSender sender) {
-		if (CHUNK_META_LOGGER.getLevel().intValue() < Level.INFO.intValue()) {
-			CHUNK_META_LOGGER.setLevel(Level.INFO);
+		if (!CHUNK_META_LOGGER.getLevel().isMoreSpecificThan(Level.INFO)) {
+			Configurator.setLevel("Chunk meta", Level.INFO);
 			sender.sendMessage(Component.text("Chunk meta logs disabled.", NamedTextColor.RED));
 		} else {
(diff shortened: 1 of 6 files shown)


plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
```diff
@@ -0,0 +1,33 @@
...
+@CommandAlias("cmc")
+public class ChunkMetaCommand extends BaseCommand {
+
+	private static boolean ENABLE_CHUNK_META_LOGS = false;
+
+	public static boolean chunkMetaLogsEnabled() {
+		return ENABLE_CHUNK_META_LOGS;
+	}
+
+	@Subcommand("chunkmeta")
```
AngrySoundTech, line 21: I don't think this is intuitive as a toggle, the command should be something like `/cmc togglechunkmetalogs` or `/cmc chunkmetalogs true`

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
```diff
@@ -0,0 +1,33 @@
...
+import co.aikar.commands.annotation.Description;
+import co.aikar.commands.annotation.Subcommand;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.bukkit.command.CommandSender;
+
+@CommandAlias("cmc")
+public class ChunkMetaCommand extends BaseCommand {
+
+	private static boolean ENABLE_CHUNK_META_LOGS = false;
```
AngrySoundTech, line 15: The command shouldn't hold what's effectively global state itself, there should be some dedicated collection of feature flags be it the config, or secondarily derived from the config

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/world/locations/chunkmeta/WorldChunkMetaManager.java
```diff
@@ -228,16 +228,22 @@ private void registerUnloadRunnable() {
 				}
 
 				if (System.currentTimeMillis() - coord.getLastUnloadedTime() > UNLOAD_DELAY) {
+					if (ChunkMetaCommand.chunkMetaLogsEnabled()) {
+						logger.info("[Chunkmeta] Unloading chunk " + coord + " - unloaded: " + coord.getLastUnloadedTime());
+					}
```
AngrySoundTech, line 233: We shouldn't be logging with guard conditions like this in modern java, we should be using the appropriate log level and a message supplier to prevent resources spent on extra checks and calculations.

This also makes the above command unnecessary.

e.g. 
```java
logger.debug(() -> "[Chunkmeta] Unloading chunk " + coord + " - unloaded: " + coord.getLastUnloadedTime())
```

AngrySoundTech (changes requested):

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/world/locations/chunkmeta/WorldChunkMetaManager.java
okx-code: I did some reasearch into this and there's not really any good options for runtime log configuration like this, and since it's temporary I just went for something easy

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
okx-code: No it shouldn't but it was the easiest option for a very accessory feature

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/commands/ChunkMetaCommand.java
okx-code: I'll change that

AngrySoundTech (approved):

PR #392, opened 2024-03-29
Refactors dependency gluing

Protonull:
This PR refactors the APIs provided by CivModCore that were intended to help hooking into soft dependencies. Now, any plugin that wishes to create a dependency glue need only do the following:

1. Ensure that CivModCore is declared within the `depend` or `softdepend` lists in your plugin's `plugin.yml`

2. Create a class that will NEVER be touched by any other code, eg:
	`src/main/java/net/civmc/example/glues/NameLayerGlue.java`
	Which must implement the `DependencyGlue` interface.

3. Create a file called `glues.properties` and with the content:
	```properties
	NameLayer:net.civmc.example.glues.NameLayerGlue
	```
	Where the key is the name of the dependency and the value the class-name of the glue.

4. Done.

That's it... the glue class will be initialised when CivModCore detects the dependency is enabled for the first time and then invoke `enable()`. If the dependency is disabled at some point, the `disable()` method is called. Any subsequent re-enabling will call `enable()`, etc.

This does mean that glues cannot themselves be API interfaces: the entire point is that they're only interacted with by CivModCore's `GlueManager` class, but this is a great way of safely registering listeners and such without fear [these kinds of issues](https://github.com/Protonull/BonkersJavacOptimisationsTest), since the class is never initialised if the dependency is not present.


Protonull:
Hmmm, starting to wonder whether this refactor is moot: the server is just the compiled output of this monorepo; there'll *NEVER* be a question over whether NameLayer, or Citadel, or JukeAlert, is missing, unless we delete those plugins from the monorepo... which is unlikely. In addition, Okx has said that we should stop pretending that we aren't the only ones who consume our APIs ([reference](https://discord.com/channels/912074050086502470/919348449407815680/1244821453707280445)), and SoundTech has said that all changes must be tested with a local docker instance ([reference](https://discord.com/channels/912074050086502470/952314898317189120/1243925834755084389)) because it replicates the server as closely as possible. With these things in mind, I'm wondering whether it'd be better to yeet this boilerplate altogether and convert all our soft-dependencies into hard-dependencies.

Protonull:
Closing due to a [lack](https://discord.com/channels/912074050086502470/919348449407815680/1255421346608054345) of [interest](https://discord.com/channels/912074050086502470/919348449407815680/1280813608875589717).

PR #393, opened 2024-03-29
Re-add standardised test output formatting

Protonull:
Re-adds [this change](https://github.com/CivMC/CivGradle/pull/2) to CivGradle. Each plugin will still need to include junit test dependencies as there doesn't seem to be a way to do so from the root `build.gradle.kts` without applying the `java-library` plugin to the root project.


Protonull:
This does make junit testing more [Maven-like](https://github.com/CivMC/CivGradle/pull/2), but the item tests fail due to changes to all the server internals. Will convert this to a draft until something better comes along.

Protonull:
Closing this in favour of what this: https://github.com/CivMC/Civ/pull/552#issuecomment-2251866756

PR #394, opened 2024-04-01
Fix Factory Overflow

MrJeremyFisher:
Fixes FactoryMod incorrectly believing that recipe output would fit in a chest because of some incorrect uses of `ItemMap`'s `fitsIn` and an issue with `fitsIn` itself.


okx-code (approved):
Thanks for fixing this

PR #395, opened 2024-04-04
Yeet canvas

Protonull:
Canvas was added in one of my chonker PRs (https://github.com/CivClassic/CivModCore/pull/73) to replace PagedGUI, a previous addition of mine (https://github.com/CivClassic/CivModCore/commit/7f64e12be0fc6d2f4c297b5267c7d316fdae004c), but it has sat unused for three years. Canvas is also a snapshot JitPack dependency in that it has NO releases, nor is maintained all that often.


okx-code (approved):
Deleting code is my favourite activity

PR #396, opened 2024-04-06
Add horse armour hack

Protonull:
This optionally allows players to craft iron, gold, and diamond horse armour like so:
image

The idea being you're taking [Leather Horse Armour](https://minecraft.fandom.com/wiki/Horse_Armor#Crafting), which is already craftable, and [plating it](https://youtu.be/E-xqDwl_UVI?feature=shared&t=68). But this can be changed.

Horse armour has the unique quality of being [invincible](https://minecraft.fandom.com/wiki/Horse_Armor#Durability), so this hack optionally reimplements durability. ~~This is somewhat hacky though since the client has no idea about this and wont show a durability bar. Open to suggestions on how this could be resolved.~~

Since horse armour is invincible, the Unbreaking enchant [has no effect](https://minecraft.fandom.com/wiki/Horse_Armor#Enchanting), so this hack optionally reimplements Unbreaking.

### DIFF
--- a/ansible/files/paper-config/plugins/SimpleAdminHacks/config.yml
+++ b/ansible/files/paper-config/plugins/SimpleAdminHacks/config.yml
@@ -452,6 +452,14 @@ hacks:
     #a list of materials players shouldn't be allowed to place at all
     noplace:
      #- GRINDSTONE Leaving this here for reference
+  HorseArmour:
+    enabled: true
+    # Register custom crafting recipes for horse armour?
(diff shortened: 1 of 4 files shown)


Protonull:
Now that CivMC is on >=1.21, I've switched this from using PDCs to using `Damageable`, though this should be switched out for [item data-components](https://github.com/PaperMC/Paper/pull/10845) at the earliest opportunity.

PR #398, opened 2024-04-09
Standardises and centralises compacted item handling

Protonull:
Moves compacted item determination out of FactoryMod and into CivModCore. This is meant to make it easier for other plugins to determine whether items are compacted without roundabout testing, effectively hard-coding predicates based on how FactoryMod is configured.


Protonull:
After the most recent commits, compacted items no longer contain the "Compacted Item" lore. That lore is instead added just prior to the item being sent over the network. This doesn't contaminate the original item, see screenshot of using `/debugging rawitem`:

image

okx-code:
I like the way you've approached this, it makes it look like compacted items should've always been done this way. I'll give it a proper review when I have the time.

Protonull:
Thanks :) I've just noticed a few little things to clean up like [this](https://github.com/Protonull/CivPlugins/blob/9f3a17d1400e1e67ac5a297a2e587cb2d5c9dbde/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/compaction/CompactionTestCommands.java#L77) and this(https://github.com/Protonull/CivPlugins/blob/9f3a17d1400e1e67ac5a297a2e587cb2d5c9dbde/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/compaction/Compaction.java#L19), but other than that it *should* be good to go, pending some verification testing.

Protonull:
Just tested this with FactoryMod. I used `/fmtest give recipe INGREDIENTS adv_xp_1` to give me all the inputs that XP 1 of the Advanced Cauldron and ran it, and it successfully produced 2 compacted emerald blocks. I then ran those through a decompactor and got 2 stacks of emerald blocks.

Protonull:
Just realised that this'll need to include a new ItemExchange rule modifier since items wont have the lore anymore to match against.

walkersgaming:
Will this also affect wordbank recipes that use compacted items?

Protonull:
> Will this also affect wordbank recipes that use compacted items?

Yeah, it probably would as it changes the hash of the item, though keep in mind that the 1.20.5 update will *probably* break most, if not all wordbank recipes, for the same reason.

PR #481, opened 2024-04-30
Eliminate Intermediate containers

AngrySoundTech:
This PR removes the intermediate step of building our own containers, instead using the base container and templating in our plugins with Ansible. This does cause the deploy to be marginally slower, but in the short term cuts some fat out of our release process and gets us moving again.


okx-code (approved):
i'm happy for you or sorry that happened

PR #497, opened 2024-05-09
Build plugins from source for ansible deployments

AngrySoundTech:
With this PR, we will no longer build our plugins just to check the jars into the repo for ansible.

Artifacts will be collected by the `:ansible:build` gradle task prior to deployment. Vendored jars are moved to `src/paper-plugins`


AngrySoundTech:
This should be good, but I didn't want to hijack the test server from 1.20 to test it too hard. We can merge it, merge it up to 1.20, and then test it and make any fixes as needed.

RedDevel2 (approved):

PR #512, opened 2024-05-12
Fix emotes

RedDevel2:
Was originally part of GSit but got seperated into it's own GEmotes plugin. This just restores functionality we already had for patreon supporters.


ansible/variables/settings.yml
```diff
@@ -31,7 +31,7 @@ settings:
     environment: Test
     minecraft:
       <<: *default_minecraft
-      civdocker_version: '2.0.0'
+      civdocker_version: 'Emotes'
```
okx-code, line 34: change

okx-code (changes requested):

ansible/variables/settings.yml
RedDevel2: woops done

Issue #93, opened 2023-12-20
Players lose XP on combat logging


AngrySoundTech:
Bug as a result of CivMC/CombatTagPlus#21

PR #535, opened 2024-05-14
Revert "Discard entities before adding NPCs"

okx-code:
This reverts commit 5d2151f1cd458b66db140586924c28cb20cb698d, from PR https://github.com/CivMC/CombatTagPlus/pull/21.

This fixes players losing their bed position as well as players losing their crafting recipes and xp (https://github.com/CivMC/Civ/issues/93), because the fact that the UUID is the same, combined with https://github.com/CivMC/Civ/blob/f525b7afeb0fa63e4cef3e48c3ec1d7dc790fc5d/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/listener/PlayerListener.java#L303, causes an interaction where the player's saved data is overriden with the NPC's data, which has *most* things copied but not the bed location. This change seems pretty broken, and was only applied recently with the 1.20 update, so I decided to roll it back until we implement a fix for everything.

Alternatively, we could make sure we are copying all of the data to the NPC, but since this has caused ~two~ three (https://github.com/CivMC/Civ/issues/93) independent issues already, it may cause more, and I think the safest option is to roll it back.

### DIFF
--- a/plugins/combattagplus-paper/src/main/java/net/minelink/ctplus/nms/NpcPlayer.java
+++ b/plugins/combattagplus-paper/src/main/java/net/minelink/ctplus/nms/NpcPlayer.java
@@ -33,7 +33,7 @@ public class NpcPlayer extends ServerPlayer {
     public static NpcPlayer valueOf(Player player) {
         MinecraftServer minecraftServer = MinecraftServer.getServer();
         ServerLevel worldServer = ((CraftWorld) player.getWorld()).getHandle();
-        GameProfile gameProfile = new GameProfile(player.getUniqueId(), NpcNameGeneratorFactory.getNameGenerator().generate(player));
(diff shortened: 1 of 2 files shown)


Protonull:
Given how fraught with bugs combat loggers have been, it might be worth looking into custom Paper patches. These issues all stem from CombatTagPlus being a plugin, and so its approach can only be to duplicate the player entity upon logout. But this has led to many, *many* bugs, including dupes, because combat-loggers are consequential: they're meant to *be* the player.

Whereas, if combat loggers were implemented through patches (or mixins), you wouldn't need to create a duplicate entity. You could just prevent the server from deleting the player-entity when they logout. There'd likely still be issues to iron out, but having the same entity to work with would be a lot easier than playing Whack-A-Mole with these combat-logger bugs.

RedDevel2 (approved):

Issue #510, opened 2024-05-12
Named/Lored items don't stack properly


RedDevel2:
Items that have data since the 1.20 update no longer stack properly. They often have slight differences in the data on them (Why)
Could have something to do with the SAH Item meta converter hack no longer working properly?

This mainly affects Essence, Fossils, and compacted items.
Some of these items no longer work in the factories related to #495

Protonull:
Right, just went about investigating this and we have an annoying problem. When [this issue](https://github.com/CivClassic/CivModCore/issues/88) was opened (on 1.16.5), setting an item's display name via `itemMeta.displayName(Component.text("Hello, World!"))` would yield the following NBT:
```
{
    "Name": '{"text":"Hello, World!"}'
}
```
Whereas, `itemMeta.setDisplayName("Hello, World!")` would yield this NBT:
```
{
    "Name": '{"text":"","extra":[{"text":"Hello, World!"}]}'
}
```

However, it seems that Minecraft has done some optimisations because components in NBT are now (in 1.20.4) looking a lot like their pre-1.13 selves, where `itemMeta.displayName(Component.text("Hello, World!"))` will yield the following NBT:
```
{
    "Name": 'Hello, World!'
}
```
And `itemMeta.setDisplayName("Hello, World!")` yields this NBT:
```
{
    "Name": '{"text":"","extra":["Hello, World!"]}'
}
```

This seems like an easy fix, right? Just do `Component.empty().append(displayName)`, except that produces this NBT:
```
{
    "Name": '{"extra":["Hello, World!"],"text":""}'
}
```

Despite these being identical JSON objects, they are not identical strings... and they are stored on items as serialised strings. I'm not sure it's possible to fix this issue. It may be worth hard-upgrading items to 1.20.4 components instead.

okx-code:
Not relevant anymore

Protonull:
> Not relevant anymore

I do think this'll likely come back to bite you sooner or later. For example, bastion ingredients (like Casings and Gearboxes) are still defined by item-meta lore, rather than by your custom-item api, so if Mojang changes how components are stored again then the problem will arise again. It's just not a problem *right now*.

PR #547, opened 2024-06-09
Bandaid fix nbt matching issues

MrJeremyFisher:
Should fix issues impacting stacking and factory use of items like essence, fossils, and compacted items. See the video below for how to trigger the update. Re-logging is equivalent to a server restart, so any inventories opened before a restart will have their contents fully converted post-restart. 

I'm working on a proper fix that will happen in "real time" but for now this will at least make the items useable again (I think! This issue has been a real headache to deal with).

Addresses #510 

[1280x720.webm](https://github.com/CivMC/Civ/assets/63616270/36407aee-0fbd-4f0f-9696-93a57882ac93)

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/ItemMetaConverterHack.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/ItemMetaConverterHack.java
@@ -4,11 +4,16 @@ import com.programmerdan.minecraft.simpleadminhacks.SimpleAdminHacks;
 import com.programmerdan.minecraft.simpleadminhacks.framework.BasicHack;
 import com.programmerdan.minecraft.simpleadminhacks.framework.BasicHackConfig;
 import net.kyori.adventure.text.Component;
+import net.md_5.bungee.api.chat.TextComponent;
+import org.bukkit.Bukkit;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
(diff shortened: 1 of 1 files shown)


MrJeremyFisher:
The underlying issue here is resolved by 1.20.6 as far as I can tell so this shouldn't be necessary anymore.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/ItemMetaConverterHack.java
```diff
@@ -51,7 +52,7 @@ public static void processItem(final ItemStack item) {
 			for (int i = 0, l = currentLore.size(); i < l; i++) {
 				final var loreLine = currentLore.get(i);
 				if (!ChatUtils.isBaseComponent(loreLine)) {
-					currentLore.set(i, Component.text().append(loreLine).build());
+					currentLore.set(i, Component.empty().append(loreLine.compact()));
```
okx-code, line 55: why not loreLine.compact() ?

MrJeremyFisher: It's what worked. I'm just going to close this for now because the item issue didn't occur in 1.20.6.

ZeroQLi:
So the plan to fix this is now to update to 1.20.6?

MrJeremyFisher:
Yeah, though I'm not sure how 1.20.6 will handle items that were created before the update. Hoping any issues regarding that will crop up in testing so I can fix them before they happen on live.

PR #552, opened 2024-06-26
Remove PsuedoServer

okx-code:
It's annoying to maintain and not actually necessary


Protonull:
It's only "not actually necessary" because the tests that need it have been commented out. If you uncomment them, you'll see why it's necessary, but PseudoServer actually needs to be updated, which is why drafted #393.

Protonull:
Perhaps a better solution would be moving it over to the test source set. It's otherwise just taking up space in the production jar when nothing other than the [commented out] tests need it.

okx-code:
I'm not extending internal Bukkit classes

okx-code:
FakePlayer frm HiddenOre also needs to go but it's a mess to keep it while also keeping cave ore generation a thing

Protonull:
> I'm not extending internal Bukkit classes

Yeah, PseudoServer has the Minestom approach [of only adding what you need], whereas extending CraftServer would be the PaperMC approach [of disabling what you don't want]. PseudoServer was added back in the 1.16.4 days (before even the move to Gradle) to avoid spinning up a whole server (and figuring out where the server files should go and when they should be cleaned up) just to test some item stuff. Now that we have [:runServer](https://github.com/jpenilla/run-task), this is *far* less of an issue.

okx-code:
mockbukkit btw

Protonull:
btw, mockbukkit != craftbukkit

Protonull:
If you intend to go through with this, you should probably remove junit and the test source set altogether too. We simply do not have well any enough managed or organised projects for it, and the stuff we need to test for, like item matching, requires a minimum level of server bootstrapping. It would probably be better to move any test stuff to `civmodcore-test` with a custom run command that starts a short-lived server instead.

Issue #496, opened 2024-05-09
Setting fuel input on factories with middle click does not work. 


RedDevel2:
Currently can only be done using macros that simulate middle clicks.

MrJeremyFisher:
I'm able to set fuel input with middleclick on a local civ instance and on the test server. Is there some error when it's tried on live?

walkersgaming:
Are you in creative when testing?

MrJeremyFisher:
Hah, thanks. Looks like this isn't something that can be fixed. The client [no longer sends](https://bugs.mojang.com/browse/MC-46584) any information to the server when an item is middle clicked in an inventory while in survival. The set fuel key will have to be changed to something that is detectable.

PR #565, opened 2024-07-16
Change fuel button to left shift

MrJeremyFisher:
Addresses #496


plugins/factorymod-paper/src/main/java/com/github/igotyou/FactoryMod/utility/IOConfigSection.java
```diff
@@ -87,7 +87,7 @@ private Clickable getIoClickable(Material adjacentType, Direction dir, String di
                 .append(Component.text(dirState.displayName).color(TextColor.color(dirState.color)))
                 .asComponent());
         ItemUtils.addComponentLore(display,
-            Component.text("L/M/R click to toggle I/F/O")
```
AngrySoundTech, line 90: LS as Left Shift is a little ambiguous here, how can we make this more understandable?

AngrySoundTech (changes requested):

plugins/factorymod-paper/src/main/java/com/github/igotyou/FactoryMod/utility/IOConfigSection.java
MrJeremyFisher: I could change it to L+LS. Or move all of them from L/R to Left/Right to be even more clear

okx-code: How about we make this three lines, one for each option so they don't need to be abbreviated?

okx-code (approved):

PR #568, opened 2024-07-17
Change README to tell users to use gradle wrapper

MrJeremyFisher:


README.md
```diff
@@ -10,7 +10,7 @@ This monorepo will eventually contain all civ projects and development
 A docker compose stack is provided to help test containers built from
 this repo. To start the stack, run the following commands:
 
-1. `gradle :ansible:build`
+1. `./gradlew :ansible:build`
```
AngrySoundTech, line 13: This is only the case for linux/unix, should be `.\gradlew.bat build` for PowerShell and I believe `gradlew.bat build` for cmd on windows?

AngrySoundTech (changes requested):

README.md
MrJeremyFisher: Sorry, missed this. Powershell and CMD on Windows 10 both accept `.\gradlew`

AngrySoundTech:
`.\gradlew :ansible:build` doesn't always work on windows, only in powershell which is isn't fully POSIX compliant. should use `gradlew.bat`

MrJeremyFisher:
Does it not work for you in cmd? 
image

okx-code:
gradlew.bat would be preferable here

AngrySoundTech (approved):

PR #570, opened 2024-07-20
Update comment on pr instead of creating new one

Huskydog9988:
Fixes issue of bot sending new comment instead of updating the existing one

### DIFF
--- a/.github/workflows/comment_on_pr.yaml
+++ b/.github/workflows/comment_on_pr.yaml
@@ -50,12 +50,13 @@ jobs:
           echo "Head SHA: $HEAD_SHA"
           echo "HEAD_SHA=$HEAD_SHA" >> "$GITHUB_ENV"
 
-      - name: Delete old comments
-        uses: izhangzhihao/delete-comment@master
+      - name: Find Comment
+        uses: peter-evans/find-comment@v3
+        id: fc
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          delete_user_name: github-actions[bot]
-          issue_number: ${{ env.PR_NUMBER }}
+          issue-number: ${{ env.PR_NUMBER }}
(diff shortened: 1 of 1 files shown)


.github/workflows/comment_on_pr.yaml
```diff
@@ -50,12 +50,13 @@ jobs:
...
+      - name: Find Comment
+        uses: peter-evans/find-comment@v3
+        id: fc
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          delete_user_name: github-actions[bot]
-          issue_number: ${{ env.PR_NUMBER }}
+          issue-number: ${{ env.PR_NUMBER }}
+          comment-author: 'github-actions[bot]'
+          body-includes: "Build Successful!"
```
AngrySoundTech, line 59: This won't find the comment if the build failed, no?

Huskydog9988: Yes? But that seems to be the behavior already? It's probably worth adding a check and saying it failed anyway.

Huskydog9988: ok so i just confirmed it, because of the conditional statement on line 13, it will never run on a failed build

okx-code (approved):

okx-code:
@Huskydog9988 seems like this change is broken so I reverted https://github.com/CivMC/Civ/actions/runs/11354153170/job/31580752772

This issue looks related and might help to fix https://github.com/peter-evans/find-comment/issues/172

Huskydog9988:
The issue is related to how the workflow gets the pr number from github actions. You can see the failure in the step before.
```log
Previous Job ID: 11354132682
Previous Suite ID: 29662588586
Artifact ID: 
Pull request Number: null
Head SHA: null
```
 I might be able to figure out a fix for this, but github actions is like black magic to me, so we'll see.

PR #573, opened 2024-07-21
Fix commands for Kira

MrJeremyFisher:
Commands through ACF which use a `Player` rather than a `CommandSender` cause ACF to add a flag that checks if the player is holding an item, which Kira doesn't want to do.


AngrySoundTech:
I'm a little wary of this many changes to commands so I'm gonna spend some time to look through this before approving

PR #578, opened 2024-07-27
Reset warmup on reinforcement update

MrJeremyFisher:
Prevents being able to instantly repair a bastion by patching the associated reinforcement. Also prevents switching groups and materials without a warmup.


okx-code (dismissed):
Should include a call to `Bastion.getBastionStorage().updated()`

AngrySoundTech (changes requested):
oops

plugins/bastion-paper/src/main/java/isaac/bastion/BastionBlock.java
```diff
@@ -299,6 +299,13 @@ public long getPlaced() {
         return placed;
     }
 
+    /**
+     * Sets how long the bastion has been in the world in ms
+     */
+    public void setPlaced(long placed) {
+        this.placed = placed;
```
okx-code, line 306: updated() call should be in here

okx-code (approved):

PR #602, opened 2024-08-20
Use same slot UUID for modifiers

MrJeremyFisher:
Fixes random UUIDs causing armor and weapons to be non-compactable


Protonull:
It may be worth just updating the Slot class to hold a UUID instead of the most and least, then accessing that UUID when needed.

MrJeremyFisher:
Yeah. I can do that when I get a chance

okx-code:
Using UUIDs in AttributeModifiers is a deprecated API, I'm going to update it to use NamespacedKeys instead.

PR #604, opened 2024-08-22
separate nether ores and add copper ore to safebreak

kickylol:


okx-code:
What is the use case for adding copper

kickylol:
personally I use copper ore as a compact way to store it since using fortune on it can give you up to 20 raw copper at a time.

okx-code (approved):

PR #607, opened 2024-08-24
Add /sidebar command to disable scoreboard

Diet-Cola:
Should be able to be used by console and players, needs testing


okx-code:
This doesn't seem to update the sidebar correctly. It disables the settings but they are still visible until it gets updated. Closing as not worth the effort for me to fix

PR #615, opened 2024-08-30
Remove MoreArrayUtils

Protonull:
Removes MoreArrayUtils, which goes *almost* entirely unused save for the double-cloning in ClonedInventory that's now fixed :P


okx-code:
Why do we have so many random utility classes wtf

Protonull:
These `More*` classes are entirely my fault :P it's a mixture of me being younger and dumber, Minecraft/Paper having fewer useful shaded dependencies, and Java having fewer static utilities. For example, [`java.lang.Math.clamp()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Math.html#clamp\(double,double,double\)) was only added in Java 21. [`java.util.List.of()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/List.html#of\(E...\)) was added in Java 9, but we only stopped compiling against Java 8 in [December 2020](https://github.com/CivClassic/style-guide/pull/10).

Issue #484, opened 2024-05-01
Suggestion: Prevent Pearled players from pearling other players.


AngrySoundTech:
We may or may not want this based on discussion, but at the moment any unpearled players' excursions into the nether require being "friendly pearled" in order to protect themselves from being pearled by pearled players, who they have hardly any defenses against.

Pearled players should be able to kill anybody, but should they be able to pearl them? All this does is necessitate the player to go free their own pearl, which is a hassle.

However, unpearled players should still be able to pearl players in the nether, even though that pearl can not leave the nether. This still keeps an element of risk or something.

Shadno-DGP:
ill throw in my 2 cents (assuming this is where im supposed to do that)
why im against pearled players pearling people boils down to

1) It allows pearled players to impact wars/conflicts directly in pvp.
During the generic war, players previously pearled were able to pearl and kite 2 estalian fighters into an RI spike in the nether, where they were held for most of the war. Similarly right now, its possible to pearl someone and put them into a spike, nether bunker, etc despite already being pearled yourself.

2) It creates an unintuitive situation for defenders of nether infra
Instead of simply going through a portal and defending infra, its almost necessary to friendly pearl yourself, travel to your infra, and then gear up to defend. This is... weird? and it punishes people _for not being pearled_, which to me is a silly incentive.

3) In combination with other factors, it makes nether infra unsustainable
If a pearled player can return with no limits or ways to stop them long term to break bastions & builds, nether infrastructure / resources are essential for exp, AND these pearled players can pearl you if you go to defend, this in combination has made the nether extremely difficult to control. Taking out one of these factors (and this seems to be the one most potentially agreeable to the admins) will make the others significantly less of a pain to deal with.

Tldr; its weird that pearled attackers can show up whenever but defenders have to friendlypearl ahead of time or be at a disadvantage / risk being spiked.

awoo-civ:
In practical terms, currently the only viable way to defend Nether infra is to have a self-pearled player logged out there 24/7 ready to log in to defend. Quite unreasonable.

drekamor:
@wingzero54 @RedDevel2 @AngrySoundTech was there any resolution on this? I can write the relevant code if you confirm the issue's status

RedDevel2:
Since kicky's pr was merged we just have to enable it to do this. I think it would make sense.

PR #619, opened 2024-09-02
Prevent pearled people from pearling others

kickylol:
Addresses #484


ansible/files/paper-config/plugins/ExilePearl/config.yml
```diff
@@ -260,6 +260,7 @@ general:
   max_pearled_alts: 1
   max_pearled_message: 'You have too many imprisoned alts'
   can_elytra_with_pearls: true
+  can_pearled_players_pearl_others: false
```
Protonull, line 263: Might be worth setting this to true, as then the feature can be merged without any actual change to the game, which can be a different discussion.

kickylol: very true, default value is now true

PR #632, opened 2024-10-11
Prevent reinforced concrete powder hardening

walkersgaming:
Prevent reinforced concrete powder hardening

Reinforced concrete powder will still harden when placed directly next to water but won't when water flows into a placed block


okx-code (changes requested):
You can use MaterialTags from paper for this instead of making your own https://jd.papermc.io/paper/1.21.1/com/destroystokyo/paper/MaterialTags.html

Also, the conditional should be !(x && y) rather than !x && !y. Although you only really need to check the new state, rather than both the old and new state.

okx-code (approved):

Issue #64, opened 2022-09-28
Prevent placing blocks on nether roof


wingzero54:
We utilize Paper's setting to give damage to kill a player if they are on the nether roof. But players still have time to place some blocks, or can use printer to place blocks above the nether roof. So we should disable block placement entirely.

AngrySoundTech:
This PR needs to be moved to the monorepo

Protonull:
> Should be implemented by a configurable per-dimension height limit

You could just piggyback off the Paper setting. It's a bit of a PITA to get to, but I was last able to access it with:
```java
((CraftWorld) world).getHandle().paperConfig.netherVoidTopDamageHeight
```

drekamor:
Opened a PR https://github.com/CivMC/Civ/pull/634

PR #634, opened 2024-10-11
Add BuildLimits hack

drekamor:
Addresses https://github.com/CivMC/Civ/issues/64

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
@@ -1,0 +1,49 @@
+package com.programmerdan.minecraft.simpleadminhacks.configs;
+
+import com.programmerdan.minecraft.simpleadminhacks.SimpleAdminHacks;
+import com.programmerdan.minecraft.simpleadminhacks.framework.SimpleHackConfig;
+import com.programmerdan.minecraft.simpleadminhacks.framework.utilities.BuildLimit;
+import org.bukkit.configuration.ConfigurationSection;
+
+import java.util.List;
+
+public final class BuildLimitsConfig extends SimpleHackConfig {
+	private boolean enabled;
+	private BuildLimit[] buildLimits;
+
+	public BuildLimitsConfig(SimpleAdminHacks plugin, ConfigurationSection base) {
+		super(plugin, base);
+	}
+	public BuildLimitsConfig(ConfigurationSection base) {
+		super(SimpleAdminHacks.instance(), base);
+	}
+
+	@Override
+	protected void wireup(ConfigurationSection config) {
+		this.enabled = config.getBoolean("enabled");
+
+		if (this.enabled) {
+			List<?> rawList = config.getList("limits");
+			if (rawList != null && rawList.size() > 0) {
+				try {
+					this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
+					plugin().getLogger().info("buildlimits enabled");
+				} catch(ArrayStoreException ase) {
+					plugin().getLogger().warning("buildlimits was enabled, but is invalid");
+					ase.printStackTrace();
+				}
+			} else {
+				plugin().getLogger().warning("buildlimits was enabled, but is missing or empty");
+			}
+		} else {
+			plugin().getLogger().info("buildlimits disabled");
+		}
+	}
+
+	public boolean isEnabled() {
+		return this.enabled;
+	}
+	public BuildLimit[] getBuildLimits() {
+		return this.buildLimits;
+	}
+}
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
@@ -19,16 +19,10 @@ public final class BuildLimitsConfig extends SimpleHackConfig {
 
 	@Override
 	protected void wireup(ConfigurationSection config) {
-		if (this.shouldEnable()) {
-			List<?> rawList = config.getList("limits");
-			if (rawList != null && rawList.size() > 0) {
-				try {
-					this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
-				} catch(ArrayStoreException ase) {
-					ase.printStackTrace();
-				}
-			}
-		}
+        List<?> rawList = config.getList("limits");
+        if (rawList != null && rawList.size() > 0) {
+            this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
+        }
 	}
 	public BuildLimit[] getBuildLimits() {
 		return this.buildLimits;
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
@@ -1,4 +1,4 @@
-package com.programmerdan.minecraft.simpleadminhacks.configs;
+package com.programmerdan.minecraft.simpleadminhacks.configs.buildlimit;
 
 import com.programmerdan.minecraft.simpleadminhacks.SimpleAdminHacks;
 import com.programmerdan.minecraft.simpleadminhacks.framework.SimpleHackConfig;
@@ -8,7 +8,6 @@ import org.bukkit.configuration.ConfigurationSection;
 import java.util.List;
 
 public final class BuildLimitsConfig extends SimpleHackConfig {
-	private boolean enabled;
 	private BuildLimit[] buildLimits;
 
 	public BuildLimitsConfig(SimpleAdminHacks plugin, ConfigurationSection base) {
@@ -20,28 +19,16 @@ public final class BuildLimitsConfig extends SimpleHackConfig {
 
 	@Override
 	protected void wireup(ConfigurationSection config) {
-		this.enabled = config.getBoolean("enabled");
-
-		if (this.enabled) {
+		if (this.shouldEnable()) {
 			List<?> rawList = config.getList("limits");
 			if (rawList != null && rawList.size() > 0) {
 				try {
 					this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
-					plugin().getLogger().info("buildlimits enabled");
 				} catch(ArrayStoreException ase) {
-					plugin().getLogger().warning("buildlimits was enabled, but is invalid");
 					ase.printStackTrace();
 				}
-			} else {
-				plugin().getLogger().warning("buildlimits was enabled, but is missing or empty");
 			}
-		} else {
-			plugin().getLogger().info("buildlimits disabled");
 		}
-	}
-
-	public boolean isEnabled() {
-		return this.enabled;
 	}
 	public BuildLimit[] getBuildLimits() {
 		return this.buildLimits;
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
@@ -1,0 +1,81 @@
+package com.programmerdan.minecraft.simpleadminhacks.framework.utilities;
+
+import org.bukkit.configuration.serialization.ConfigurationSerializable;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class BuildLimit implements ConfigurationSerializable {
+	private String world;
+	private String type;
+	private int min_y;
+	private int max_y;
+
+	public BuildLimit (String world, String type, int min_y, int max_y){
+		this.world = world;
+		this.type = type;
+		this.min_y = min_y;
+		this.max_y = max_y;
+	}
+
+	public String getWorld() {
+		return this.world;
+	}
+	public String getType() {
+		return this.type;
+	}
+	public int getMin_y() {
+		return this.min_y;
+	}
+	public int getMax_y() {
+		return this.max_y;
+	}
+
+	public void setWorld(String world) {
+		this.world = world;
+	}
+	public void setType(String type) {
+		this.type = type;
+	}
+	public void setMin_y(int min_y) {
+		this.min_y = min_y;
+	}
+	public void setMax_y(int max_y) {
+		this.max_y = max_y;
+	}
+
+	@NotNull
+	@Override
+	public final Map<String, Object> serialize(){
+		Map<String, Object> result = new LinkedHashMap<>();
+
+		result.put("world", this.world);
+		result.put("type", this.type);
+		result.put("min_y", this.min_y);
+		result.put("max_y", this.max_y);
+
+		return result;
+	}
+	@NotNull
+	public static BuildLimit deserialize(@NotNull Map<String, Object> args){
+		String world = "world";
+		String type = "altitude";
+		int min_y = 0;
+		int max_y = 0;
+
+		world = (String) args.get("world");
+		type = (String) args.get("type");
+
+		if(type.equals("altitude")){
+			if(args.containsKey("min_y")){
+				min_y = (Integer) args.get("min_y");
+			}
+			if(args.containsKey("max_y")){
+				max_y = (Integer) args.get("max_y");
+			}
+		}
+
+		return new BuildLimit(world, type, min_y, max_y);
+	}
+}
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
@@ -9,14 +9,14 @@ import java.util.Map;
 public class BuildLimit implements ConfigurationSerializable {
 	private String world;
 	private String type;
-	private int min_y;
-	private int max_y;
+	private int minY;
+	private int maxY;
 
-	public BuildLimit (String world, String type, int min_y, int max_y){
+	public BuildLimit (String world, String type, int minY, int maxY){
 		this.world = world;
 		this.type = type;
-		this.min_y = min_y;
-		this.max_y = max_y;
+		this.minY = minY;
+		this.maxY = maxY;
 	}
 
 	public String getWorld() {
@@ -25,11 +25,11 @@ public class BuildLimit implements ConfigurationSerializable {
 	public String getType() {
 		return this.type;
 	}
-	public int getMin_y() {
-		return this.min_y;
+	public int getMinY() {
+		return this.minY;
 	}
-	public int getMax_y() {
-		return this.max_y;
+	public int getMaxY() {
+		return this.maxY;
 	}
 
 	public void setWorld(String world) {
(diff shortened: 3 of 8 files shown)


plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
...
+		return this.max_y;
+	}
+
+	public void setWorld(String world) {
+		this.world = world;
+	}
+	public void setType(String type) {
+		this.type = type;
+	}
+	public void setMin_y(int min_y) {
```
Protonull, line 41: `setMin_y` ?

drekamor: What's wrong?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
+package com.programmerdan.minecraft.simpleadminhacks.framework.utilities;
+
+import org.bukkit.configuration.serialization.ConfigurationSerializable;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class BuildLimit implements ConfigurationSerializable {
```
Protonull, line 9: If this class will only be used by the BuildLimits hack, it could be made an inner-class of the hack. That's just a personal preference though.

drekamor: SoundTech OK'ed this so :man_shrugging:

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
Protonull: Just seems like an odd naming convention to have both camel and snake case. It's typical for getters in Java to use camel casing (eg: `setMinY`) or otherwise to have the field public if you want to retain the name.

drekamor: Oh, I used `y_min` in the config so it just got carried over to the var names. Now that I look at it, it does feel a bit weird

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
Protonull: Okay :+1:

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
awoo-civ: Yeah `setMinY` is correct, you never wanna make exceptions to name casing.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
```diff
@@ -0,0 +1,49 @@
...
+		super(plugin, base);
+	}
+	public BuildLimitsConfig(ConfigurationSection base) {
+		super(SimpleAdminHacks.instance(), base);
+	}
+
+	@Override
+	protected void wireup(ConfigurationSection config) {
+		this.enabled = config.getBoolean("enabled");
+
```
awoo-civ, line 24: nit: remove this empty line

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+
+		if(!withinLimits(loc)){
+			e.setCancelled(true);
+			player.sendMessage(Component.text().content("You cannot place blocks here").color(NamedTextColor.RED));
+		}
+	}
+
+	private boolean withinLimits(Location loc){
+		BuildLimit[] limits = config.getBuildLimits();
+
```
awoo-civ, line 49: nit: remove this empty line

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+		plugin.registerListener(this);
+	}
+
+	@EventHandler(priority = EventPriority.HIGHEST)
+	public void onBlockPlaceEvent(BlockPlaceEvent e){
+		if(!config.isEnabled()) return;
+
+		Player player = e.getPlayer();
+		Location loc = e.getBlock().getLocation();
+
```
awoo-civ, line 40: nit: remove this empty line

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+		if(!withinLimits(loc)){
+			e.setCancelled(true);
+			player.sendMessage(Component.text().content("You cannot place blocks here").color(NamedTextColor.RED));
```
awoo-civ, line 43: Move the declaration of `player` to above this line or just use `e.getPlayer()` directly.

awoo-civ:
Style: 
- `if(limit.getType().equals("altitude")){` -> `if (limit.getType().equals("altitude")) {`
- `public void onEnable(){` -> `public void onEnable() {`
- `public static BuildLimitsConfig generate (` -> `public static BuildLimitsConfig generate(`

etc.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
...
+	private String world;
+	private String type;
```
awoo-civ, line 11: `type` should ideally be an enum instead of a string.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
...
+
+		result.put("world", this.world);
+		result.put("type", this.type);
+		result.put("min_y", this.minY);
+		result.put("max_y", this.maxY);
+
+		return result;
+	}
+	@NotNull
+	public static BuildLimit deserialize(@NotNull Map<String, Object> args){
```
awoo-civ, line 61: public static BuildLimit deserialize(@NotNull Map<String, Object> args) {
		int minY = 0;
		int maxY = 0;
		String type = "altitude";
		type = (String)args.get("type");
		if (type.equals("altitude")) {
			if (args.containsKey("min_y")) {
				minY = (Integer)args.get("min_y");
			}
			if (args.containsKey("max_y")) {
				maxY = (Integer)args.get("max_y");
			}
		}

		String world = "world";
		world = (String)args.get("world");

		return new BuildLimit(world, type, minY, maxY);
	}

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
okx-code: There doesn't seem to be a `type` other than altitude. Is this necessary?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
```diff
@@ -0,0 +1,49 @@
...
+	public BuildLimitsConfig(SimpleAdminHacks plugin, ConfigurationSection base) {
+		super(plugin, base);
+	}
+	public BuildLimitsConfig(ConfigurationSection base) {
+		super(SimpleAdminHacks.instance(), base);
+	}
+
+	@Override
+	protected void wireup(ConfigurationSection config) {
```
okx-code, line 22: Put this all in onEnable and then you don't have to worry about checking if the hack is enabling in the config. Also, don't bother logging that the hack is enabled, and let errors throw, since that will disable the hack (or rethrow with a nicer message if you want)

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
```diff
@@ -0,0 +1,49 @@
...
+					plugin().getLogger().info("buildlimits enabled");
+				} catch(ArrayStoreException ase) {
+					plugin().getLogger().warning("buildlimits was enabled, but is invalid");
+					ase.printStackTrace();
+				}
+			} else {
+				plugin().getLogger().warning("buildlimits was enabled, but is missing or empty");
+			}
+		} else {
+			plugin().getLogger().info("buildlimits disabled");
```
okx-code, line 39: You don't need to log if a hack is enabled or disabled, that's done already.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
...
+		this.minY = minY;
+	}
+	public void setMaxY(int maxY) {
+		this.maxY = maxY;
+	}
+
+	@NotNull
+	@Override
+	public final Map<String, Object> serialize(){
+		Map<String, Object> result = new LinkedHashMap<>();
```
okx-code, line 51: Why LinkedHashMap here? Is insertion order important?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,81 @@
...
+		return this.type;
+	}
+	public int getMinY() {
+		return this.minY;
+	}
+	public int getMaxY() {
+		return this.maxY;
+	}
+
+	public void setWorld(String world) {
```
okx-code, line 35: are the setters necessary?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public void onEnable(){
+		plugin.registerListener(this);
+	}
+
+	@EventHandler(priority = EventPriority.HIGHEST)
+	public void onBlockPlaceEvent(BlockPlaceEvent e){
+		if(!config.isEnabled()) return;
```
okx-code, line 36: unnecessary

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+
+	public static BuildLimitsConfig generate (
+			final @NotNull SimpleAdminHacks plugin,
+			final @NotNull ConfigurationSection config){
+		return new BuildLimitsConfig(plugin, config);
+	}
+
+	@Override
+	public void onEnable(){
+		plugin.registerListener(this);
```
okx-code, line 31: unnecessary

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
okx-code: `Component.text(String, TextColor)` could be used to make this more succint, but you don't have to do it that way

okx-code (changes requested):

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/BuildLimitsConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+	@Override
+	protected void wireup(ConfigurationSection config) {
+		this.enabled = config.getBoolean("enabled");
+
+		if (this.enabled) {
+			List<?> rawList = config.getList("limits");
+			if (rawList != null && rawList.size() > 0) {
+				try {
+					this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
```
Protonull, line 29: Instead of try-catching an assignment, you could use an iterator to check that each element is an instance of `BuildLimit`, which then lets you print nice warnings for each offending entry.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
Protonull: This one actually is necessary: it's only BasicHack that registers itself as a listener upon `onEnable`

okx-code: goddamnit i hate hacks

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
```
Protonull, line 33: Needs an `onDisable` override here to remove this listener, then the `if(!config.isEnabled()) return;` checks are not necessary since the event will only ever be called when the hack is enabled, because the listeners will only ever be registered upon hack enable.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
drekamor: I think they are. There isn't a great lot of documentation on serialisations so I used what I could find

okx-code: Why?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
drekamor: SoundTech asked to add `type` because he wants to add others (e.g radius limit) in the future

drekamor: Making it an enum would mean adding yet another useless for anything else class. Can't make it a subclass because it needs to be accessed from different packages.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
drekamor: `Listener#unregisterListener` is depricated

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
awoo-civ: I do not understand the concern with adding another class, even if it was necessary which I don't think it is? Would defining a public enum in an already existing class not work just fine?

In any case, using a String is quite bad form as:
1. It can be anything - easy for bugs to sneak in
2. Finding out what it can be is near impossible
Enums exist for the exact purpose of fixing these issues.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/BuildLimits.java
Protonull: Use [`HandlerList.unregisterAll(this);`](https://github.com/CivMC/Civ/blob/c6278bd01319bbf55c415113efd89ba3cad61f2a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/BasicHack.java#L23), which is what BasicHack uses.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
```diff
@@ -0,0 +1,36 @@
...
+	public BuildLimitsConfig(SimpleAdminHacks plugin, ConfigurationSection base) {
+		super(plugin, base);
+	}
+	public BuildLimitsConfig(ConfigurationSection base) {
+		super(SimpleAdminHacks.instance(), base);
+	}
+
+	@Override
+	protected void wireup(ConfigurationSection config) {
+		if (this.shouldEnable()) {
```
okx-code, line 22: This shouldEnable() call is still unnecessary and will cause unexpected behaviour if something happens like the hack is disabled in the config, but then is enabled by a command.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/configs/buildlimit/BuildLimitsConfig.java
```diff
@@ -0,0 +1,36 @@
...
+			List<?> rawList = config.getList("limits");
+			if (rawList != null && rawList.size() > 0) {
+				try {
+					this.buildLimits = rawList.toArray(new BuildLimit[rawList.size()]);
+				} catch(ArrayStoreException ase) {
+					ase.printStackTrace();
```
okx-code, line 28: This throwing is fine as it will disable the hack, which will be more visible to the administrator. Just printing the error like this will cause strange behaviour as the hack will still be enabled but not initialised properly.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/framework/utilities/BuildLimit.java
```diff
@@ -0,0 +1,50 @@
...
+	@Nullable
+	public static BuildLimit deserialize(@NotNull final Map<String, Object> args){
+		String world = (String) args.get("world");
+
+        String type = (String) args.get("type");
+        LimitType limitType = LimitType.ALTITUDE;
+        try {
+            limitType = LimitType.valueOf(type);
+        } catch (IllegalArgumentException e) {
+            return null;
```
okx-code, line 34: Throwing is fine here as the exception will be handled and printed in a clearer rather than discarded as is what happens here. See `org.bukkit.configuration.serialization.ConfigurationSerialization#deserializeViaMethod`

okx-code (changes requested):

okx-code (approved):

PR #635, opened 2024-10-11
enforce lf file endings

Huskydog9988:
Have git enforce lf endings


Protonull:
This would force `gradlew.bat` to use LF endings, making it unusable on Windows.

Huskydog9988:
forgot about that, will add an exemption

okx-code (approved):

PR #638, opened 2024-10-13
Fix bastions acting as sponge and stacking incorrectly

okx-code:


plugins/hiddenore-paper/src/main/java/com/github/devotedmc/hiddenore/util/FakePlayer.java
```diff
@@ -2253,6 +2253,11 @@ public Spigot spigot() {
         return null;
     }
 
+    @Override
```
Diet-Cola, line 2256: hiddenore stuff in bastion pr?

okx-code: yeah it was necessary to get this to build. But I deleted this class anyway so it doesn't matter

PR #645, opened 2024-10-20
Increase item frame chunk limits

okx-code:


RedDevel2:
This just updates the /chunklimits list. You'll have to change it in the Insights config too.

RedDevel2 (approved):

PR #654, opened 2024-11-01
Fix a bug

drekamor:
@RedDevel2


okx-code:
Could we disable canStorageTeleport instead?

drekamor:
I don't know why we have it so I've made a solution that wouldn't affect it in any way :man_shrugging:

okx-code:
It should be safe to removed based on the commit history  https://github.com/CivMC/Civ/commit/0a9b5b4568b65cb3083ef24fc26fb4135f98f381

drekamor:
Well, then you can do that. Discovering the cause of the bug was the hard part here. Took 5 minutes to write that code anyway

okx-code:
Alright

PR #656, opened 2024-11-10
Show pearl health value instead of percent

AngrySoundTech:
This PR Does not include any changes to pearl cost.

It only changes the way we display pearl health, showing the actual health value instead of a percentage.
Additionally, it updates the "Cost per day" to reflect how much of this health the pearl is actually using.

Fresh pearl:
image

Refueling preview:
image

Refueling output:
image

Four year old pearl:
image

### DIFF
--- a/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/ExilePearl.java
+++ b/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/ExilePearl.java
@@ -124,13 +124,6 @@ public interface ExilePearl {
     int getHealth();
 
     /**
-     * Gets the pearl health percent value
-     *
-     * @return The health percent value
-     */
-    Integer getHealthPercent();
-
-    /**
      * Sets the pearl health value
      *
      * @param health The health value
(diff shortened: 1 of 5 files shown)


okx-code:
Does 1 Health = 1 Essence? If so we should show it as "Essence:" rather than "Health:"

AngrySoundTech:
> Does 1 Health = 1 Essence? If so we should show it as "Essence:" rather than "Health:"

Not currently. Right now, one essence repairs 12 health, and 24 health is consumed per day (one every hour).

As the pearl ages, essence repairs less and less health during crafting. Moving that part to when the fuel is consumed instead of refueled is part of my next goal of uncapping pearl health.

awoo-civ:
> Not currently.

Oh! That is extremely confusing then.

AngrySoundTech:
> 1. Health and Cost per day lines should be next to each other
> 
>     2. Show "(days remaining: 200)", to spare the need for the player to count
> 
>     3. It seems unreasonable for the cost to be able to be 75 with the health at 336, that requires refueling every 4.5 days

1. I agree, I'll move that in this PR.

2. Days remaining is planned in a follow-up PR

3. It's strange yes, but that's math works with the cost multiplier.

AngrySoundTech:
Health moved above cost, days left will be there too in that grouping.

image

RedDevel2 (approved):

PR #670, opened 2024-12-10
Add ItemExchange receipts

Protonull:
This implements [this suggestion](https://discord.com/channels/912074050086502470/1313232152263000156/1313232152263000156). It adds a new setting to `/config` (defaulted to `true`) which lets users decide whether to receive receipts for their ItemExchange purchases. This setting can be overridden-to-true by the `ReceiptModifier` modifier.

The receipt contains the following information:

- `Purchaser`
- `Timestamp` of purchase (pretty printed)
- `Shop block`: the block-location of the relay/shop-chest punched to use the shop
- `Shop hash`: the `.hashCode()` of the block-location of the container hosting the exchange itself, and where the input item(s) will be deposited into, and the output item(s), if any, are withdrawn from.
- The "input" chat-display for the exchange.
- The "output" chat display for the exchange, if any.
- Any footer text / slogan set on the `ReceiptModifier`.
- A receipt-unique UUID.

This adds the following commands:

- `/ies receipt`: This will toggle having a `ReceiptModifier` on the held exchange rule.
- `/ies receipt alwaysprint`: This will toggle whether to always print receipts, regardless of customer preference, on the held exchange rule.
- `/ies receipt footer clear`: Clears any set footer-text on the held exchange rule.
- `/ies receipt footer set <text>`: Sets footer-text on the held exchange rule.

In addition, receipts are given the custom-model id of `-44_054_986` (just a random int) so that modders can easily assign it a custom texture.

---

While this does ["dupe"](https://discord.com/channels/912074050086502470/952314898317189120/1314411805149233243) items, receipts are just stone buttons (like exchange rules themselves), which are perhaps one of the cheapest items in the game and can already be "duped" with `/iec`.

### DIFF
--- a/ansible/files/paper-config/plugins/ItemExchange/config.yml
+++ b/ansible/files/paper-config/plugins/ItemExchange/config.yml
@@ -141,3 +141,6 @@ shopRelay:
     - GREEN_STAINED_GLASS_PANE
     - RED_STAINED_GLASS_PANE
     - BLACK_STAINED_GLASS_PANE
+
+receipts:
+  enabled: false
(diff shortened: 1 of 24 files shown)


Protonull:
> Make this disabled by default and I'll merge it

Done

PR #671, opened 2024-12-10
Faction permissions support

OwainX:
AI created code changes, only basic testing done so far.  Requires review.

This change creates faction.[groupName] permissions for namelayer groups, adds the permission to a user when they are added to a group and removes the permission when they are removed.  This change will allow MythicMobs mobs to target based on namelayer group membership.


okx-code:
This would be much better served as a LuckPerms context, not a permission imo

AngrySoundTech:
Rejecting this. We review and accept changes from humans who have put in the effort, not ethically dubious untested and unreviewed changes generated with AI.

PR #675, opened 2024-12-27
Update DestinationCommand.java to suggest OneDest structure

Zalvvv:
With this feature when writing /dest the command helper would suggest in the following order: 
!
+,-  +,+  -,+  or  -,- (quadrant)
Impendia Moloka  Lyrean  MtA  Medi  Arctic  AP  Deluvia  Founders  Proxima  Icenia  or  Karydia (region)

OneDest follows a tree structure that I don't know how to apply to the commandCompletion. It would be preferable so anybody puts a region that is not in the input quadrant. Onedest tree can be found here: https://github.com/OneDest/data/blob/master/tree.txt

Though, if not possible/too complicated, I think that my solution is already helpful


Protonull:
This should be configurable rather than hard coded, so that it can be changed similarly to RandomSpawn cities

Zalvvv:
I totally, agree but i don't know how to do it :c

AngrySoundTech:
Rejecting this. Admins will not enforce or endorse any one routing system over the other, as they are player made constructs/concepts within the game

PR #708, opened 2025-01-10
Mostly fix JSON notes

MrJeremyFisher:
Notes and books will now print the correct (same as book and plate) content. Annoyingly, before this, they never printed the correct content due to the `.replaceAll("(?<!§x(§[\\da-f]){0,5})" + ChatColor.BLACK, ChatColor.GRAY.toString())` which incorrectly replaced some colours. This means that the removal of that line results in correct colouring but makes printed items after this change different from printed items before it. Obviously this would change wordbank recipes so it probably merits some discussion. I've attached images of what items look like now without the replaceAll, with it, and what they looked like before 1.20.5
image
image
image


okx-code:
Why remove the replaceall?

okx-code (changes requested):
Why remove the replaceall?

MrJeremyFisher:
It was fixing a fix that wasn't necessary anymore. In its current state it works as it should. With the replaceall it was incorrectly replacing formatting characters as can be seen in my original comment.

okx-code:
Why is the fix not necessary anymore? I don't want existing json notes to be printed differently if they don't have to be

PR #729, opened 2025-01-15
fix mobs spawning/stuck outside simulation distance

RedDevel2:
Some mobs are not de-spawning correctly. or are spawning beyond simulation distance and get stuck.

### DIFF
--- a/ansible/files/paper-config/config/paper-world-defaults.yml
+++ b/ansible/files/paper-config/config/paper-world-defaults.yml
@@ -173,28 +173,44 @@ entities:
     despawn-range-shape: ELLIPSOID
     despawn-ranges:
       ambient:
-        hard: 96
+        hard:
+          horizontal: 32
+          vertical: 96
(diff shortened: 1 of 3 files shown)


okx-code (changes requested):
https://www.paper-chan.moe/paper-optimization/#despawn-ranges recommends a minimum despawn range of 36, even if the simulation distance is 3. mob spawn range should also be 3

RedDevel2:
Should simulation distance be bumped to 4 then? because despawn ranges don't work in the case of mob groups spawning on the edge of the chunk, causing some being beyond the chunk-border and getting stuck in limbo.

okx-code:
no, we can make simulation distance of 3 work. Changing it to 4 would also increase the number of loaded chunks a lot and therefore lag.

RedDevel2:
3 it is then.

okx-code (approved):

PR #735, opened 2025-01-22
Reattempt ModSupport hack

Protonull:
This is a reattempt of #582 that's better written and with more features. As mentioned in the previous PR, this leverages plugin messages to send additional information that the Minecraft protocol lacks. This should help with civ-mods and mods updated to work with civ.


okx-code:
I don't think this is the right direction to go for working around rules limitations, it's better to change the rules.

Protonull:
The problem is that, in my opinion, it's very difficult if not impossible to reword the rules to include these things but also don't accidentally include Baritone.

PR #738, opened 2025-01-24
Add restart announcements

### LINKED ISSUE #473: Add restart warning

Huskydog9988:
Very basic restart announcement system. Uses cron expressions to schedule when specific messages are to be sent. In the future I want to make this to handle both dogfacts, and notify players when the server is going to be taken down for an update.  Fixes #473


wingzero54:
Have you tested this locally will the full server setup?

Huskydog9988:
Yes

plugins/announcements-velocity/src/main/java/net/civmc/announcements/ProxyListener.java
```diff
@@ -0,0 +1,5 @@
+package net.civmc.announcements;
+
+public class ProxyListener {
```
okx-code, line 3: what does this do?

plugins/announcements-velocity/src/main/resources/config.yml
```diff
@@ -0,0 +1,10 @@
+
+# messages must be in json format
+# https://minecraft.tools/en/json_text.php
+scheduledAnnouncements:
+  - cron: "45 1 * * *"
+    message: "{\"text\":\"Restart in 15 minutes\",\"color\":\"gold\"}"
```
okx-code, line 6: is it possible to support minimessage here?

plugins/announcements-velocity/src/main/java/net/civmc/announcements/ProxyListener.java
Huskydog9988: guess i forgot to remove this class, is the recommended place for event listeners on the proxy though iirc. Will need this for the future changes i want, but forget which ones need it.

plugins/announcements-velocity/src/main/resources/config.yml
Huskydog9988: i guess? but id probably do it in another pr if your okay with that

okx-code: would rather it be in one pr

okx-code (changes requested):
This doesn't work, it sends one message every second for the entire minute an announcement is supposed to be sent.

image

Huskydog9988:
That's strange, I don't remember it doing that. Could you send me the config you used?

okx-code:
I just used the example config and updated it to the next minute

```
scheduledAnnouncements:
  - cron: "22 3 * * *"
    message: "{\"text\":\"Restart in 15 minutes\",\"color\":\"gold\"}"
  - cron: "21 3 * * *"
    message: "{\"text\":\"Restart in 5 minutes\",\"color\":\"gold\"}"
  - cron: "20 3 * * *"
    message: "<rainbow>Restart</rainbow> in 1 minute"
    minimessage: true
```

Huskydog9988:
That's what I figured but just wanted to be sure. Will look into it in a bit.

plugins/announcements-velocity/src/main/java/net/civmc/announcements/AnnouncementsPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+
+            // get last time a cron *should* have run
+            executionTime.lastExecution(now).ifPresent(lastExecution -> {
+                // Check if we should execute, and if con already executed at that time
+                if (executionTime.isMatch(now) && !lastExecution.equals(lastExecutionTimes.get(cron))) {
+                    // Execute and update the last execution time
+                    var announcement = scheduledAnnouncements.get(cron);
+                    server.sendMessage(announcement.message);
+                    lastExecutionTimes.put(cron, lastExecution);
+                    if (announcement.showTitle) {
```
okx-code, line 104: This throws an NPE if it's null

okx-code (changes requested):

PR #745, opened 2025-01-27
Allow adding blocks to safebreak

MrJeremyFisher:
Allows players to add whatever blocks they want to safebreak. Keeps support for the default blocks in the config.


okx-code:
Safebreak is only for silk touched blocks

MrJeremyFisher:
Why?

okx-code:
Safe ore break exists so that you can be el cheapo and use a non-silk pickaxe for vein mining because ores pop in so quickly. It's not there to stop you accidentally mining bookshelves or whatever

MrJeremyFisher:
I'd agree except bookshelves (and sea lanterns, and glowstone) are already included in the config on live.

okx-code:
Mistake imo

MrJeremyFisher:
Then why not let people just add whatever they want? That way you don't have to think about the horrors of people adding glass or whatever

PR #746, opened 2025-01-28
Case-insensitive Commands

MrJeremyFisher:
Reintroduces the case-insensitive behaviour of commands from pre-1.19.1. Does not change the case of command arguments.

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CaseInsensitivity.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CaseInsensitivity.java
@@ -1,0 +1,20 @@
+package com.programmerdan.minecraft.simpleadminhacks.hacks.basic;
+
+import com.programmerdan.minecraft.simpleadminhacks.SimpleAdminHacks;
+import com.programmerdan.minecraft.simpleadminhacks.framework.BasicHack;
+import com.programmerdan.minecraft.simpleadminhacks.framework.BasicHackConfig;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.player.PlayerCommandPreprocessEvent;
+
+public class CaseInsensitivity extends BasicHack {
+
+    public CaseInsensitivity(SimpleAdminHacks plugin, BasicHackConfig config) {
+        super(plugin, config);
+    }
+
+    @EventHandler
+    public void onCommandPreProcess(PlayerCommandPreprocessEvent event) {
+        String[] splitMessage = event.getMessage().split(" ", 2);
+        event.setMessage(splitMessage[0].toLowerCase() + " " + splitMessage[1]);
+    }
+}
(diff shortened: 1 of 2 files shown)


Protonull:
Immediate concerns here are:

- Are you absolutely certain there'll always be >1 array elements when split?

- How will this effect commands like `/nlcreategroup`? Will group names be lowercased?

MrJeremyFisher:
Thanks. I may be stupid. Commands with arguments are fine as that's the purpose of the split.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CaseInsensitivity.java
```diff
@@ -0,0 +1,24 @@
...
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.player.PlayerCommandPreprocessEvent;
+
+public class CaseInsensitivity extends BasicHack {
+
+    public CaseInsensitivity(SimpleAdminHacks plugin, BasicHackConfig config) {
+        super(plugin, config);
+    }
+
+    @EventHandler
```
Protonull, line 15: I'd argue that the priority should be set to LOWEST so that the transformation is done as early as possible. Any event handlers thereafter would then be deciding based on the amended command.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CaseInsensitivity.java
```diff
@@ -0,0 +1,24 @@
...
+    public void onCommandPreProcess(PlayerCommandPreprocessEvent event) {
+        String[] splitMessage = event.getMessage().split(" ", 2);
+        if (splitMessage.length == 1) {
```
Protonull, line 18: Since you're doing length checks, I'd advocate doing an `indexOf()` instead since that's arguably has a more obvious control flow (I can only speak to myself, but I had to look up the documentation for that version of split to see what the second parameter would do).

```java
final String command = event.getMessage();
final int spaceIndex = command.indexOf(" ");
if (spaceIndex == -1) {
	return command.toLowerCase();
}
return command.substring(0, spaceIndex).toLowerCase() + command.substring(spaceIndex);
```

MrJeremyFisher: I'm not sure about this one. To me the split makes intuitive sense vs. what you've got there. I'll leave it up to the admins on whether or not they care.

Diet-Cola:
Only long term concern with this is how future civ servers will 100% not realise this is a thing

Protonull:
Honestly, I'm surprised this isn't an option somewhere in Minecraft/Bukkit/Spigot/Paper

okx-code:
I don't think we should be rewriting how commands work like this. If mojang changed it, let it be

MrJeremyFisher:
I agree but consider that it'd be nice to have and we already override a bunch of Mojang changes

PR #747, opened 2025-01-29
Update death messages

Protonull:
Literally only made this because Vulcan1776 is killing people with "Baked Potato" as opposed to the "Jacket Potato" (`en_gb` locale). This PR also yeets the kill-broadcast formatting setting in favour 'show weapon' boolean setting.

### DIFF
--- a/plugins/civchat2-paper/src/main/java/vg/civcraft/mc/civchat2/listeners/KillListener.java
+++ b/plugins/civchat2-paper/src/main/java/vg/civcraft/mc/civchat2/listeners/KillListener.java
@@ -1,8 +1,12 @@
 package vg.civcraft.mc.civchat2.listeners;
 
-import java.util.Optional;
+import java.util.Objects;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.TextComponent;
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.format.Style;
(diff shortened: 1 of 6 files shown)


plugins/civchat2-paper/src/main/java/vg/civcraft/mc/civchat2/listeners/KillListener.java
```diff
@@ -1,94 +1,142 @@
...
-                .orElse(false);
-            if (!hasWordBank) {
-                displayName = String.format("with a %s", itemName);
-            } else {
-                String killMessageFormat = settingsMan.getKillMessageFormat(killer.getUniqueId()).simpleDescription;
-                if (killMessageFormat.isBlank()) {
-                    displayName = String.format("%s%s", killMessageFormat, displayName);
-                } else {
-                    displayName = String.format("%s %s", killMessageFormat, displayName);
+        final Component killMessage; {
```
okx-code, line 72: huh???

Protonull: What's the problem?

Protonull: Since I cannot read your mind, I can only assume you found the scoped building of the kill message confusing, so I have moved it into its own function 👍

Issue #706, opened 2025-01-10
damage from a player is ignored if you are already in a damage tick


RedDevel2:
Damage dealt by players is nullified when the opponent is already in a damage tick from fire,poison,cactus etc.

This means having fire aspect on your sword or infecting your opponent with poison is actually disadvantageous. This is counterintuitive and opposite of what it should be, an advantage.

Confirmed on the pvp server, but likely also present on live)

okx-code:
I think this is only a problem on pvp

PR #680, opened 2025-01-05
Fix packet listener weirdness

MrJeremyFisher:
Duplicate of #676 on oldfinale. Also fixes paper repo url.

PR #750, opened 2025-02-01
Combat fix

MrJeremyFisher:
Addresses #706. Sorry for commit history, I accidentally based the branch off of #680.

### DIFF
--- a/plugins/finale-paper/src/main/java/com/github/maxopoly/finale/combat/CombatUtil.java
+++ b/plugins/finale-paper/src/main/java/com/github/maxopoly/finale/combat/CombatUtil.java
@@ -3,11 +3,8 @@ package com.github.maxopoly.finale.combat;
 import com.github.maxopoly.finale.Finale;
 import com.github.maxopoly.finale.combat.event.CritHitEvent;
 import com.github.maxopoly.finale.combat.knockback.KnockbackStrategy;
-import com.github.maxopoly.finale.misc.knockback.KnockbackConfig;
-
 import java.util.Iterator;
 import java.util.List;
-
 import net.minecraft.core.Holder;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.network.protocol.game.ClientboundSetEntityMotionPacket;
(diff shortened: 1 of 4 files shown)


plugins/finale-paper/src/main/java/com/github/maxopoly/finale/combat/CombatUtil.java
```diff
@@ -113,15 +105,16 @@ public static void attack(ServerPlayer attacker, Entity victim) {
...
+                    if (itemstack.is(ItemTags.SWORDS)) {
                         shouldSweep = true;
                     }
                 }
 
                 float victimHealth = 0.0F;
 
                 Vec3 victimMot = victim.getDeltaMovement();
-                if (shouldDamage) {
+                boolean flag4 = victim.hurtOrSimulate(damagesource, damage);
```
okx-code, line 116: what is the point of calling this function twice?

plugins/finale-paper/src/main/java/com/github/maxopoly/finale/combat/CombatUtil.java
```diff
@@ -73,8 +67,6 @@ public static void attack(ServerPlayer attacker, Entity victim) {
                 shouldDamage = f2 > 0.9f;
                 damage *= 0.2F + f2 * f2 * 0.8F;
                 f1 *= f2;
-            } else if (victim instanceof LivingEntity living) {
-                shouldDamage = living.invulnerableTime <= living.invulnerableDuration / 2;
```
okx-code, line 77: This is a line of code that took me several hours to write. Unfortunately, this is necessary because of the way that Minecraft handles knockback and damage. Removing this just introduces the bug where you can deal knockback with no cooldwon. I think that Paper should be fixing this in 1.21.3 so that may be a better way.

okx-code:
This PR doesn't fix the issue.

okx-code:
Yes, the knockback needs to work as well.

MrJeremyFisher:
Ok. I can fix that I think.

okx-code:
Paper already fixes this in 1.21.3 anyway, I'd rather we updated to that.

MrJeremyFisher:
Are we not already on 1.21.3?

okx-code:
Oh yeah my test server is on 1.21.1 though lol

okx-code:
Ah yeah we are, but the knockback issue remains

okx-code:
See https://github.com/PaperMC/Paper/pull/11599

MrJeremyFisher:
Hm. I kind of doubt any paper changes here would necessarily help us as we completely override the vanilla (or paper) method that handles this. I'll take a better look at it and the knockback issue when I have the time.

PR #752, opened 2025-02-02
Add a barrel recipe to carpentry factory

RedDevel2:


okx-code:
Actually it should be more like 16 -> 20, this is still net negative with the charcoal cost

okx-code (approved):

PR #754, opened 2025-02-04
Add ItemUpdater interface

Protonull:
This creates a standardised and easy-to-use way to update items.

---

You can create a class that implements some or all of the predefined listeners:
```java
package uk.protonull.civ.example;

import org.bukkit.inventory.ItemStack;
import org.jetbrains.annotations.NotNull;
import vg.civcraft.mc.civmodcore.ACivMod;
import vg.civcraft.mc.civmodcore.inventory.items.updater.listeners.UpdatePlayerItemsOnJoin;

public final class ExamplePlugin extends ACivMod {
    @Override
    public void onEnable() {
        super.onEnable();
        registerListener(new ItemUpdateListener());
    }

    private static final class ItemUpdateListener implements UpdatePlayerItemsOnJoin {
        @Override
        public void updateItem(final @NotNull ItemStack item) {
            // if (item.needsUpdating()) {
            //     item.updateSomehow();
            // }
        }
    }
}
```

---

Or you can have a static method and cast it to a predefined listener:
```java
package uk.protonull.civ.example;

import org.bukkit.inventory.ItemStack;
import org.jetbrains.annotations.NotNull;
import vg.civcraft.mc.civmodcore.ACivMod;
import vg.civcraft.mc.civmodcore.inventory.items.updater.listeners.UpdatePlayerItemsOnJoin;

public final class ExamplePlugin extends ACivMod {
    @Override
    public void onEnable() {
        super.onEnable();
        registerListener((UpdatePlayerItemsOnJoin) ExamplePlugin::updateItem);
    }

    private static void updateItem(final @NotNull ItemStack item) {
        // if (item.needsUpdating()) {
        //     item.updateSomehow();
        // }
    }
}
```

---

The ItemMetaConverterHack hack has been updated to the former, except its even more concise as basic hacks register themselves as event listeners. This system comes in useful when wanting to support legacy items. For example, this would be useful if and when ItemExchange switches from NBT to PDC.

### DIFF
--- a/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/custom/CustomItemsUpdater.java
+++ b/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/custom/CustomItemsUpdater.java
@@ -5,16 +5,20 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import org.bukkit.Bukkit;
+import org.bukkit.NamespacedKey;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataType;
 import org.bukkit.plugin.java.JavaPlugin;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import vg.civcraft.mc.civmodcore.inventory.items.custom.migrations.ItemMigrations;
+import vg.civcraft.mc.civmodcore.CivModCorePlugin;
 import vg.civcraft.mc.civmodcore.inventory.items.updater.ItemUpdater;
 import vg.civcraft.mc.civmodcore.inventory.items.updater.listeners.DefaultItemUpdaterListeners;
+import vg.civcraft.mc.civmodcore.inventory.items.updater.migrations.ItemMigration;
+import vg.civcraft.mc.civmodcore.inventory.items.updater.migrations.ItemMigrations;
 
 public abstract class CustomItemsUpdater implements ItemUpdater {
     protected final Logger logger = LoggerFactory.getLogger(getClass());
@@ -41,7 +45,7 @@ public abstract class CustomItemsUpdater implements ItemUpdater {
     ) {
         Objects.requireNonNull(customKey);
         if (createIfAbsent) {
-            return this.targets.computeIfAbsent(customKey, (_k) -> new ItemMigrations(customKey));
+            return this.targets.computeIfAbsent(customKey, CustomItemMigrations::new);
         }
         return this.targets.get(customKey);
     }
@@ -64,10 +68,33 @@ public abstract class CustomItemsUpdater implements ItemUpdater {
     }
 
     /**
-     * This is how this class determines whether an item is a custom item. You can just override with <i>just</i> a call
-     * to {@link vg.civcraft.mc.civmodcore.inventory.items.custom.CustomItem#getCustomItemKey(org.bukkit.inventory.ItemStack)},
-     * but you can also include some logic that allows support for pre-{@link vg.civcraft.mc.civmodcore.inventory.items.custom.CustomItem}
-     * custom items.
+     * <p>This is how this class determines whether a given item is a custom item. You can just override this with a simple
+     * call to {@link vg.civcraft.mc.civmodcore.inventory.items.custom.CustomItem#getCustomItemKey(org.bukkit.inventory.ItemStack)},
+     * but you may want to add additional logic to support legacy custom items.</p>
+     *
+     * <pre>{@code
+     * // Example
+     * @Override
+     * public @Nullable String getCustomKeyFrom(@NotNull ItemStack item) {
+     *     if (CustomItem.getCustomItem(item) instanceof final String customKey) {
+     *         return customKey;
+     *     }
+     *     if (item.getType() == Material.ENDER_EYE
+     *         && ExampleUtils.hasPlainDisplayName(item, "Player Essence")
+     *         && ExampleUtils.hasPlainLoreLine(item, "Activity reward used to fuel pearls")
+     *     ) {
+     *         return "player_essence";
+     *     }
+     *     if (item.getType() == Material.BONE_BLOCK
+     *         && ExampleUtils.hasPlainDisplayName(item, "City Bastion")
+     *         && ExampleUtils.hasPlainLoreLine(item, "City bastions block reinforcements and elytra")
+     *     ) {
+     *         return "city_bastion";
+     *     }
+     *     // etc
+     *     return null;
+     * }
+     * }</pre>
      */
     protected abstract @Nullable String getCustomKeyFrom(
         @NotNull ItemStack item
@@ -99,5 +126,36 @@ public abstract class CustomItemsUpdater implements ItemUpdater {
     ) {
         Bukkit.getPluginManager().registerEvents(DefaultItemUpdaterListeners.wrap(updater), plugin);
         return updater;
+    }
+}
+
+final class CustomItemMigrations extends ItemMigrations {
+    private static final NamespacedKey VERSION_KEY = new NamespacedKey(JavaPlugin.getPlugin(CivModCorePlugin.class), "item_version");
+
+    public CustomItemMigrations(
+        final @NotNull String customKey
+    ) {
+        // This is a deliberate 0th migration that ensures that any item
+        // being migrated has a custom-item key and item version.
+        this.migrations.put(0, (ItemMigration.OfItem) (item) -> {
+            CustomItem.setCustomItemKey(item, customKey);
+            setMigrationVersion(item, 0);
+        });
+    }
+
+    @Override
+    public @Nullable Integer getMigrationVersion(
+        final @NotNull ItemStack item
+    ) {
+        // This is a readonly PDC
+        return item.getPersistentDataContainer().get(VERSION_KEY, PersistentDataType.INTEGER);
+    }
+
+    @Override
+    public void setMigrationVersion(
+        final @NotNull ItemStack item,
+        final int version
+    ) {
+        item.editPersistentDataContainer((pdc) -> pdc.set(VERSION_KEY, PersistentDataType.INTEGER, version));
     }
 }
--- a/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/updater/migrations/ItemMigration.java
+++ b/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/updater/migrations/ItemMigration.java
@@ -1,4 +1,4 @@
-package vg.civcraft.mc.civmodcore.inventory.items.custom.migrations;
+package vg.civcraft.mc.civmodcore.inventory.items.updater.migrations;
 
 import net.minecraft.core.component.DataComponentHolder;
 import org.bukkit.craftbukkit.inventory.CraftItemStack;
--- a/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/updater/migrations/ItemMigrations.java
+++ b/plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/updater/migrations/ItemMigrations.java
@@ -1,33 +1,15 @@
-package vg.civcraft.mc.civmodcore.inventory.items.custom.migrations;
+package vg.civcraft.mc.civmodcore.inventory.items.updater.migrations;
 
 import java.util.Map;
 import java.util.Objects;
 import java.util.TreeMap;
-import org.bukkit.NamespacedKey;
 import org.bukkit.inventory.ItemStack;
-import org.bukkit.persistence.PersistentDataType;
-import org.bukkit.plugin.java.JavaPlugin;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.Range;
-import vg.civcraft.mc.civmodcore.CivModCorePlugin;
-import vg.civcraft.mc.civmodcore.inventory.items.custom.CustomItem;
 
-public class ItemMigrations {
-    public static final NamespacedKey VERSION_KEY = new NamespacedKey(JavaPlugin.getPlugin(CivModCorePlugin.class), "item_version");
-
+public abstract class ItemMigrations {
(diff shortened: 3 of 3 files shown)


Protonull:
This would also be useful for the ["Increase the durability of fishing rods, shears, flint and steel, and carrot on a stick"](https://discord.com/channels/912074050086502470/1266682338646036502/1266682338646036502) suggestion, and the ["Add horse armour hack"](https://github.com/CivMC/Civ/pull/396) PR, and any other change that involves rebalancing items somehow.

okx-code:
This is something that could prove quite useful, but I'm not sure about the way it's implemented. Perhaps there could be a setting reserved only for admins that would change whether items are updated, and items would be updated on all events, instead of having to inherit specific interfaces, which is kind of weird

Protonull:
Well, the idea with using interfaces this way is to avoid as much as possible creating an encapsulated black-box system (like ChunkMeta), which must be poked and prodded to work how you want. The only thing this requires is an `ItemUpdater` implementation, where you are encouraged to make this a public-static method somewhere, so that it can be used by the event handlers, but also callable by, say, a command. It also means that each plugin can choose the methods of update, so particularly expensive updaters are not running each and every time a chunk is loaded. I could add a default interface that extends all event handlers to make this easy?

Protonull:
Was asked by Okx in VC to shift to a DataFixer-esque style migrations, similar to ManagedDatasource, with an API resembling (pseudocode):
```java
CustomItem.registerMigration("meteoric_iron_pickaxe", version: 2, migration: item -> item.meta(meta -> meta.setMaxDurability(999999)))
```

While I have implemented this, I have done so as a layer of abstraction on top of the existing `ItemUpdater` interface:
```java
package uk.protonull.civ.example;

import io.papermc.paper.datacomponent.DataComponentTypes;
import org.bukkit.plugin.java.JavaPlugin;
import vg.civcraft.mc.civmodcore.inventory.items.updater.impl.CustomItemUpdater;

public final class ExamplePlugin extends JavaPlugin {
    @Override
    public void onEnable() {
        final CustomItemUpdater updater = CustomItemUpdater.init(this, "meteoric_iron_pickaxe");
        updater.migrations.registerDataMigration(1, (data) -> data.set(DataComponentTypes.MAX_DAMAGE, 999_999));
    }
}
```

This is because I do not believe the requested system was fit for purpose. It's my continued position that this system *should* work for vanilla items, even if it's not used that way, because there may be suggestions in future that are accepted.

I do not believe this to be unreasonable speculation, as while the [fishing rods, shears, flint and steel, and carrot on a stick](https://discord.com/channels/912074050086502470/1266682338646036502/1266682338646036502) and the [horse armour](https://github.com/CivMC/Civ/pull/396) suggestions have not [yet] been accepted, the [make emeralds food](https://discord.com/channels/912074050086502470/1298376852905136229/1298376852905136229) suggestion has. How would one do that if this system only works with custom items?

In addition, Okx seems amenable to the idea of updating ItemExchange away from NBT and onto [PDCs](https://jd.papermc.io/paper/1.21.3/org/bukkit/persistence/PersistentDataContainer.html), but how would that work when ItemExchange buttons are not considered custom items by CivModCore?

For this item updater system to have any meaningful use, it MUST be able to handle non-custom items, if only to update them to custom items. This is extremely easy to do, just see the `CustomItemUpdater` class to see how it's done.

okx-code (commented):
I like the implementation now, but please delete DataComponentHolder. It's just speculation.

Protonull:
> please delete DataComponentHolder. It's just speculation.

Weeeell, the stated reason in the documentation for why it's there is the polite version. It's there as a form of type limitation, ie, I only want to let people doing a data-component migration to have access to its data components... because giving people access to the full item will *inevitably* lead to really bad habits.

The biggest example of this that's absolutely rampant in civ-code is `ItemUtils.setDisplayName()` and `ItemUtils.addLore()` (etc) which are used in succession, like this piece of FactoryMod code:
image
Where each of those `ItemUtils` calls deserialises the item's meta, does a single thing, then re-serialises it onto the item. This is extraordinarily inefficient.

The problem is that items do not have an equivalent `DataComponentHolder` interface, so the only other way to give access to data components is just by passing the whole item into the migration methods. I'm worried that people will start modifying the items in ways outside of data components, such as the item's material or amount, or even its meta. The first in that list has pretty significant implications up to and including an entire reset of the item's meta. And while [`ItemStack#setType()`](https://jd.papermc.io/paper/1.21.3/org/bukkit/inventory/ItemStack.html#setType(org.bukkit.Material)) is now deprecated, that's never stopped anyone before. Just run a deprecation check on this monorepo and see how many warnings you get.

Protonull:
Is there anything else you'd like changed?

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/events/EventUtils.java
```diff
@@ -1,11 +1,21 @@
...
 import java.util.Objects;
+import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerInteractEvent;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.plugin.java.JavaPlugin;
 import org.jetbrains.annotations.NotNull;
 
 public final class EventUtils {
+    /** Convenience shortcut */
+    public static void registerListener(
```
okx-code, line 12: you don't need this

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/inventory/items/updater/impl/CustomItemsUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+        final @NotNull String customKey,
+        final boolean createIfAbsent
+    ) {
+        if (createIfAbsent) {
+            return this.targets.computeIfAbsent(customKey, (_k) -> new ItemMigrations(this.versionKey));
+        }
+        return this.targets.get(customKey);
+    }
+
+    public void removeMigrationsFor(
```
okx-code, line 63: why would there be a removeMigrationsFor method? Also we don't really need two classes (overhead doesn't matter) for custom items

Protonull: > why would there be a removeMigrationsFor method?

It's there to support general reloadability. For example, SimpleAdminHacks: should there be a plugin-level `CustomItemsUpdater` that hacks use, or should each hack register its own? Hacks are designed to be as reloadable as possible. It doesn't hurt to have the method there even if you don't intend to use it.

plugins/civmodcore-paper/src/main/java/vg/civcraft/mc/civmodcore/events/EventUtils.java
Protonull: It's not *needed*, sure, but it's a nice shortcut for civ plugins that don't extend `ACivMod`, like CastleGates, CombatTagPlus, ExilePearl, HiddenOre, RandomSpawn, and VoidWorld.

okx-code: It's not even convenient though

Protonull: That's because Java doesn't have extension methods D:

Protonull:
> The problem is that items do not have an equivalent `DataComponentHolder` interface

Seems like PaperMC has gotten around to adding a [`DataComponentHolder`](https://jd.papermc.io/paper/1.21.5/io/papermc/paper/datacomponent/DataComponentHolder.html) interface, but it's on 1.21.5

Protonull:
With the last few commits, I've began working turning [this](https://github.com/CivMC/Civ/blob/a677e69a062f17b3f66af1fe0884d9a712ac841b/plugins/itemexchange-paper/src/main/java/com/untamedears/itemexchange/rules/ExchangeRule.java#L244-L263) (converting the inline display-name and lore criteria into modifiers) into a proper migration. However, this has required more helper infrastructure than expected.

For example, say that Player Essence gets its `CustomItem` update, so now it has the `"player_essence"` custom key. Now say that you want to change the name from Player Essence back to MemeMana, you could just do:
```java
getMigrationsFor("player_essence").registerMetaMigration(1, (meta) -> {
    meta.displayName(Component.text("MemeMana"));
});
```
But this would override any anvil-set names since it's not checking the previous name before setting the new one, and I'd hazard a guess that, by virtue of standard human-nature laziness, that most migrations would be like that. And that's not okay. This has particular implications with the updates happening effectively randomly: when people login or open inventories. A bad migration could propagate over time, increasing the cost of a revert. And depending on the item, the issue might not be noticed for quite some time.

I am honestly considering abandoning this item updater system (though you'd still be free to merge it if you wish) in favour of a standalone executable that would load region and player files and migrate every item all at once. Migrations can therefore be targeted, done during a daily restart, tested, and reverted if it causes any issues.

okx-code:
There was an itemName method added recently, does that work better with anvil names compared to using displayName?

Protonull:
> There was an itemName method added recently, does that work better with anvil names compared to using displayName?

Oh yeah, that's explicitly why it was added. And to be fair, newer custom items like your Meteoric Iron items use item names, but older custom items like bastions, their ingredients, the DebugWand (SAH), repair kits, trim fragments, fossils, etc, are still using display names and lore. I don't think these items should stop working because someone renamed them in an anvil, but they would. Nor do I think such things should be removed via lazy migration.

The display name issue is just one example too, another could be Finale balancing. It wasn't all that long ago that axes would have literally hundreds of attributes on them because, anytime you opened an inventory, any axe in that inventory would get another damage attribute *added* to it. This probably allowed for a save-corruption dupe glitch, though I don't think it was used. Anyway, to be fair to the devs, the modern attribute API didn't exist at the time so they had to do it via raw NBT, so it's less of a laziness issue. Then again, it was a known issue for quite some time before being fixed. Regardless, it's an example of how civ has done item-migrations in the past and how that's gone wrong.

And ultimately, changing names and lore will likely be what 99% of migrations are, and it's an easy bet that this will be done as lazily as possible, and that worries me.

okx-code:
well, for meteoric, most migrations will need to be adding/removing blocks it can break, as well as lore. Also they can't be renamed in an anvil anyway.

Protonull:
I still plead that display information (names, lore, models, etc) should be done at network time, similar to what ViaBackwards does when translating new items to older clients. Civ shouldn't need permanent lambdas held in memory for the rest of time to fix a typo in lore. And given how so many custom items (and compacted items) are matched by their lore:
getMigrationsFor("player_essence").registerMetaMigration(1, (meta) -> {
    meta.lore(List.of(Component.text("Activity reward used to fuel pearls and factories")));
});
Whoops, now you just deleted everyone's compacted Player Essence. This *WILL* happen, it's inevitable. And since compacted items are not custom items, compacted items being lore-based cannot be fixed with the migrations system.

okx-code:
There should be a way to register a migration so that it can 'fail', or only update the lore and version if the lore exactly matches what is expected

Protonull:
> There should be a way to register a migration so that it can 'fail'

Sure, that's something I did with that [standardised compaction PR](https://github.com/CivMC/Civ/pull/398/files#diff-bb544b51b11f84f6a263c021aead920fe586eef126ae3e700e5aeee47d8ab411R137-R175) a while back but figured that an abstract migrations system couldn't be as granular. I'll see what I can do about adding result types.

That being said, the only way to *"only update the lore and version if the lore exactly matches what is expected"* without immediately slamming back into the issue of laziness is with hyper-specific premade migrations:
```java
getMigrationsFor("player_essence").registerMigration(1, PremadeMigrations.replaceExactPlainLoreLine(
    PremadeMigrations.LoreLine.ANY,
    "Activity reward used to fuel pearls",
    "Activity reward used to fuel pearls and factories"
));
```

While that can be done to some extent, more likely the kind of safety we both want would look like:
```java
getMigrationsFor("player_essence").registerMetaMigration(1, (meta) -> {
    final List<Component> lore = new ArrayList(meta.lore());
    boolean changed = false;
    for (final var iter = list.listIterator(); iter.hasNext();) {
        if (ExampleUtils.matchesPlainLine(iter.next(), "Activity reward used to fuel pearls")) {
            iter.set(Component.text("Activity reward used to fuel pearls and factories"));
            changed = true;
        }
    }
    if (!changed) {
        return MigrationResult.FAIL;
    }
    meta.lore(lore);
    return MigrationResult.SUCCESS;
});
```

But that's just not going to happen, at least not without significant oversight. Laziness will just result in migrations looking like:
```java
getMigrationsFor("player_essence").registerMetaMigration(1, (meta) -> {
    // Add "and factories" to lore
    meta.lore(List.of(Component.text("Activity reward used to fuel pearls and factories")));
    return MigrationResult.SUCCESS;
});
```

And to be fair to them, that's not an unreasonable migration... why shouldn't a migration be able to re-set the lore? Do migrations really need to be git patches? The whole problem here is that civ still relies on lore as identifiers for a number of custom items, and for compacted items, so re-setting the lore potentially wipes out those lore-identifiers. But this wouldn't matter if those identifiers were PDC values ([something a PaperMC emphatically recommended](https://discord.com/channels/289587909051416579/1342377752774443008/1369003262589534441)), and if the display information is added at network time, obviating the need for any migration. We wouldn't even need ProtocolLib for this, we could make it a [Velocity](https://papermc.io/software/velocity) plugin that gives an item configured lore if it has a particular custom-item key or is tagged as compacted.

Also, let's put the lore-as-identifier problem aside, meaning that it is perfectly safe to just re-set the lore entirely if you wanted to, that means your migrations could look like:
```java
final var migrations = getMigrationsFor("player_essence");
migrations.registerMetaMigration(1, (meta) -> meta.setLore("Activity reward used to fuel pearls and factores"));
migrations.registerMetaMigration(2, (meta) -> meta.setLore("Activity reward used to fuel pearls and factories")); // fix typo
migrations.registerMetaMigration(3, (meta) -> meta.setLore("Activity reward used to fuel pearls, factories, and elytras"));
migrations.registerMetaMigration(3, (meta) -> meta.setLore("Activity reward used to fuel pearls, factories, elytras, and city cores"));
```
Those previous migrations (1, 2, 3) are completely redundant, they are wasted CPU cycles. You could just remove them and the result would be the same... and which point, why even have a migrations system? You said yourself that meteoric migrations would be adding or removing blocks. Do you really need separate migrations to add some blocks here and remove some there, instead of collapsing it down to a clear-and-re-set?

Ultimately, the migration system exists because item updates need to be done carefully, but they only need to be done carefully because civ continues to do custom items in the most foolish, outdated way possible, and because of admin stubbornness (no offence) to remove the problem entirely and add lore at network time. Remove these artificial problems and the need for migrations crumbles away.

Issue #157, opened 2022-09-03
Should pearls also show where they were captured?


Protonull:
Each pearl contains information about the circumstances of the pearling like who is pearled, when they were pearled, and who pearled them. It would be good, in my opinion, to also include where they were pearled. For clarity, I don't mean the current location of the pearl, but where the pearled-player died when they became pearled. This could add an interesting dynamic regarding jurisdiction and evidence, etc. Just an idea, it's not urgent or a demand.

okx-code:
I'm closing this out as something that we won't fix. It could be revisited in the future, but it's not within our capacity to work on right now.

PR #756, opened 2025-02-06
Implements pearl-capture locations

Protonull:
This adds where the pearl was originally captured to the pearl and `/ep showall` as suggested in #157 with the idea that this will create emergent gameplay, particularly *'diplomatic discussions'* over jurisdiction.

### DIFF
--- a/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/storage/MySqlStorage.java
+++ b/plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/storage/MySqlStorage.java
@@ -6,9 +6,15 @@ import com.devotedmc.ExilePearl.PearlLogger;
 import com.devotedmc.ExilePearl.config.Document;
 import com.devotedmc.ExilePearl.config.MySqlConfig;
 import com.google.common.base.Preconditions;
+import it.unimi.dsi.fastutil.ints.Int2ObjectArrayMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+import java.util.List;
+import java.util.Set;
+import net.kyori.adventure.util.Index;
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.World;
+import org.jetbrains.annotations.NotNull;
 import vg.civcraft.mc.civmodcore.dao.ConnectionPool;
 
 import java.sql.*;
@@ -140,6 +146,7 @@ class MySqlStorage implements PluginStorage {
 
             applyMigration0001();
             applyMigration0002();
+            applyMigration0003();
 
             if (config.getMigratePrisonPearl()) {
                 migratePrisonPearl();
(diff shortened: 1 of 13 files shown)


Diet-Cola:
Potentially worth adding a command to set the capture location of a pearl for admin usage just to keep consistent with other aspects like date, type, killer etc

Protonull:
> Potentially worth adding a command to set the capture location of a pearl for admin usage

Done, lemme know what you think of the command. ExilePearl's custom command system is rather icky :P

Protonull:
Something that might need to be investigated is what I presume to be a bugfix wrt return locations. If you look at [`MySqlStorage#loadAllPearls()`](https://github.com/CivMC/Civ/pull/756/files#diff-bb6cedfb8d4cefe0ff076825a2c1528d4b6971882b854bcd073af1a6c3a1979fR279-R296), it didn't load a pearl's return location from a MySQL database, which [is what CivMC uses](https://github.com/CivMC/Civ/blob/91210183d38623278c7121093e5fee0706ffb61f/ansible/files/paper-config/plugins/ExilePearl/config.yml#L244-L257). It don't think the return location is *ever* retrieved from the database, save from the addition in this PR...

Protonull:
> Why would the return location be needed to be saved across restart? If anything shouldn't the pearled player return to the end if they are summoned during a restart?

Well, presumably it's better not to attempt a return during the shutdown process because I don't believe there's a way to detect the initiation of a shutdown. There's `JavaPlugin#onDisable()` but I think all players have been kicked by that point. And also because return the player, ie, teleporting them, may require chunk loading, perhaps even world loading, to ensure the teleport is completed safely, which is probably unwise during a shutdown process which may be killed early.

But I also don't necessarily see anything wrong with summonings persisting between restarts. I don't think restarts should be considered as a probation-ending event, or even a canon-event. So if the summoning can persist between restarts, then the return location should also persist.

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/storage/MySqlStorage.java
```diff
@@ -202,75 +212,152 @@ private boolean applyMigration0002() {
...
     }
 
+    private boolean applyMigration0003() {
+        try (final Connection conn = this.db.getConnection()) {
+            try (final PreparedStatement stmt = conn.prepareStatement("ALTER TABLE exilepearls ADD CONSTRAINT unique_pearl_id UNIQUE (pearl_id);")) {
+                stmt.execute();
+            }
+            try (final PreparedStatement stmt = conn.prepareStatement(
+                """
+                create table if not exists ep_capturelocations(
```
okx-code, line 222: can we use the same table instead of creating a new table?

okx-code (changes requested):

plugins/exilepearl-paper/src/main/java/com/devotedmc/ExilePearl/storage/MySqlStorage.java
Protonull: Could do, I was only following the example of return locations, which are their own table. But if you'd prefer it as part of the `exilepearls` table, that's fine.

PR #759, opened 2025-02-15
Reattempt KeyedUtils pruning

Protonull:
This is a reattempt of #400, which prunes various parts of `KeyedUtils`, particularly the `fromParts()` method since the relevant `NamespacedKey` constructor [was de-deprecated](https://github.com/PaperMC/Paper/issues/7753). `KeyedUtils` now fulfils its intention of filling in the gaps of keys, rather than attempting to provide pseudo null safety.


okx-code (changes requested):
Can we just delete KeyedUtils entirely? It doesn't seem to do much

Protonull:
> Can we just delete KeyedUtils entirely? It doesn't seem to do much. dittoKey isn't used.

It isn't used *yet*, I was going to use it for the ItemExchange NBT->PDC update :(

PR #761, opened 2025-02-15
Remove MoreArrayUtils

Protonull:
Somewhat of a reattempt of #615, but prunes `MoreArrayUtils` in lieu of outright deleting it. What remains is a remade `computeElements()` into `cloneArrayAndElements()` since that's its only usage (`ClonedInventory#cloneInventory()`). It was originally removed along with the entire class in #615 as the documentation and/or implementation of `Inventory` would clone the items in `getContents()`/`setConents()` regardless, so why duplicate the work? However, I think this has since changed, or perhaps I was mistaken. So cloning is necessary to ensure that any changes made to a `ClonedInventory` do not contaminate the original inventory. Unfortunately, I can't seem to find a deep-clone function anywhere, so `MoreArrayUtils` remains.


okx-code:
Could we just inline the deep clone method?

Protonull:
> Could we just inline the deep clone method?

Done

PR #763, opened 2025-02-15
Add damage_per_run option

Diet-Cola:


okx-code (changes requested):
Is it possible to make a wrapper clase for this? So you can do something like `new DamagePerRunRecipe(recipe, 1)` instead of having to change every single recipe class

PR #774, opened 2025-02-23
Make pearl exiled on date use UTC

Huskydog9988:
This brings the date more in line with the civ convention of having every date and time in game UTC based.


Protonull:
There are multiple instances of `SimpleDateFormatter` in ExilePearl. Perhaps we should have a static instance somewhere? Otherwise different places will have different times (such as pearl lore vs `/ep showall`)

Gjum:
> Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

If you want to reuse a static instance, make sure it's only ever accessed from the main thread. You could also use a factory to create identical instances for each usage.

Protonull:
Using a factory method is probably the best idea

Huskydog9988:
I switched to a factory because I just want it to work

PR #775, opened 2025-02-24
warn user when they reach max essence

Huskydog9988:


Diet-Cola:
Only advice is maybe should we drop the excess essence? Still a good change

Huskydog9988:
I'd love for that feature too, but iirc it was rejected by the admins.

okx-code (approved):

PR #779, opened 2025-02-27
add public heliodor config

Huskydog9988:
realized that the heliodor plugin was failing to start because it lacked a config


okx-code:
We don't do this for other plugins eg hiddenore, I don't want to risk this breaking deployments

PR #806, opened 2025-07-11
Update map links, add civmap.com

Gjum:


walkersgaming:
Looks like you forgot to switch the click event links 😄

PR #811, opened 2025-07-30
Improve banstick's proxy ip collection

Huskydog9988:
Changes:
- Switched from client9's 6 year old datacenter list, to growlfm's maintained list
- Added another source of public proxy ips (PROXY-List)
- Fixed scraping of Free Proxy List's site

If I were in charge, I'd also switch on iphub, as that's disabled for some reason

Forgot to mentioned but the pvp server and main server interfere with each other's ratelimits. I'd fix it, but that could be (sorta) temporarily solved with adjusting the config values, or with a rearchitecting (which would probably never be merged).


RedDevel2:
(Think this was merged with the Zorweth update)

Huskydog9988:
I just did a quick browse and it doesn't seem it was

Huskydog9988:
@okx-code please review

PR #816, opened 2025-08-19
Rename `Enemy Bastion` -> `Foreign Bastion`, `Ally Bastion` -> `Friendly Bastion`

tybug:
https://discord.com/channels/912074050086502470/1406920055488512070

Untested, so I may have missed some strings.


Liam-DeVoe:
is this a rejection of this suggestion, or? it was very well received on discord: https://discord.com/channels/912074050086502470/1406920055488512070 and I think is substantially more intuitive to new players

okx-code:
Yes, I think "foreign" conveys an idea of nationality that doesn't really exist. I would be amenable to "Hostile Bastion", but I think Enemy Bastion is also fine.

Issue #813, opened 2025-08-09
The Wool dying factory is missing from the /fm menu.


RedDevel2:
Not sure why, it's just not there on any page.

PR #829, opened 2025-09-04
Fix & Improvement - The Wool dying factory is missing from the /fm menu. #813 

byter-man:
Tested Realistic Biome commands and fm command  ✅
Scrollbox behaviour looks good ✅
Wool Dying factory now appears in fm list ✅


okx-code (changes requested):
This fixes the /fm factory list but it breaks the recipe list in any given factory, since scrollOffset is just interpreted as itemsPerPage now
image

PR #833, opened 2025-09-20
Make kira auth global

### LINKED ISSUE #831: /discordauth only works on main, not on pvp or on mini
Unable to run the command on these other servers, it just shows as unknown command. This is confusing behavior to new players, and may rapidly become much more relevant if people are stuck in the queue for longer. Being able to have these people auth without sitting in queue for however long could help a lot with retention.

Huskydog9988:
Makes kira's discord auth a global command instead of requiring users to only be on main to use it. One current problem with this setup is that kira has no idea what the civ username of a player is, only their mojang username. 

~~Requires #738~~
Fixes #831


okx-code:
Oh you fixed the commit nice

Huskydog9988:
Yea i copied the wrong one

PR #834, opened 2025-09-20
fix tp command using player nicknames

Huskydog9988:


Huskydog9988:
nvm apparently this solution is via?

okx-code:
Going to change names on the proxy instead

PR #836, opened 2025-09-23
very basic void queue in minestom

Huskydog9988:


okx-code:
We will revisit in the future if we need this

PR #841, opened 2025-09-26
added webhook, changed filter to avoid false flags

ghost:


plugins/civchat2-paper/src/main/java/vg/civcraft/mc/civchat2/webhook.java
```diff
@@ -0,0 +1,41 @@
...
+            String webhookUrl = "";
+            String content = "🚨 **Filtered message**\n**Sender:** " + sender + "\n**Message:** " + message;
+
+            // Escape JSON special characters
+            content = content.replace("\"", "\\\"").replace("\n", "\\n");
+
+            String json = "{\"content\": \"" + content + "\"}";
+
+            URL url = new URL(webhookUrl);
+            HttpURLConnection conn = (HttpURLConnection) url.openConnection();
```
okx-code, line 20: Didn't Java add a more modern HTTP API?

PR #847, opened 2025-10-01
Update config.yml

hakr-civ:
Changing line 129 from `true` to `false` - allowing lava to effect water in a more vanilla manner.


hakr-civ:
discussed previously - change to effect how water and lava interact

PR #849, opened 2025-10-04
Potential ItemExchange fix

Protonull:
This uses a number of strategies to try to weed out these weird ItemExchange bugs, whose causes remain unknown:

- Created an InventoryAccessor interface to specifically target an inventory's "contents" or "storage contents". This should avoid potential bugs coming from Minecraft attempting to equip a pickaxe as their helmet.

- Created an ItemStash record whose sole purpose is to guarantee anyone with an instance of it that you have a not-null, not-empty list of items which are themselves not-null, not-empty, and have valid amounts. All the "safely" inventory APIs have been switched over to this, meaning that it's now impossible to trade an empty-collection of items. This should avoid such potential bugs, or at least put a nice big stacktrace in the console.

- Those "safely" inventory APIs have had their inventory-cloning behaviour removed in favour of the caller doing it themselves. This should help with any potential issues caused by ClonedInventory being in control over whether a cloned-inventory can clone, etc, which may avoid situations where an inventory is cloned but not committed back when it should.

- Those "safely" inventory APIs now clone the item before passing it into Bukkit's .addItem() and .removeItem() inventory APIs. The javadoc for these functions say that the given item may be mutated by the inventory. This may have had implications where a failed transaction contaminated a subsequent successful transaction.


plugins/itemexchange-paper/src/main/java/com/untamedears/itemexchange/commands/TestCommand.java
```diff
@@ -0,0 +1,188 @@
...
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.PlayerInventory;
+import org.bukkit.util.BlockIterator;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import vg.civcraft.mc.civmodcore.inventory.InventoryAccessor;
+import vg.civcraft.mc.civmodcore.inventory.InventoryUtils;
+
+@CommandAlias(TestCommand.ALIAS)
+public final class TestCommand extends BaseCommand {
```
okx-code, line 25: Does this command have any permissions attached?

Protonull:
The fix for this is apparently due to `Bukkit.createInventory()` for player inventories not creating identically-functioning inventories, hence items being put into equipment slots. An [issue](https://github.com/PaperMC/Paper/issues/13154) has been made for this but the maintainers seem unwilling to fix this issue or offer alternatives. This PR proposes a fix by taking advantage of `CraftInventoryCustom`, something that has been stable [for 13 years](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/inventory/CraftInventoryCustom.java) (see blame). However, this is seen as unacceptable.

The only permitted paths ahead are reinventing the wheel for adding/remove items to/from inventories (but as ItemStack arrays) or creating an extremely brittle case for cloned player inventories where it just creates a 4x9 chest inventory and hoping that Minecraft/Paper never expands the player storage-contents (eg: to include the offhand). Because at least these aren't using perhaps the oldest and most stable Bukkit internal. Using Bukkit internals is not ideal but are rife throughout civ-code but apparently this single additional use is beyond the pale.

Someone else can deal with this... I'm done.

Issue #278, opened 2024-02-21
Renamed items showing in kill messages.


walkersgaming:
Back when I made [this PR](https://github.com/CivMC/CivChat2/pull/14) to improve global kill messages to buff the use of Wordbank, it was agreed that renamed items shouldn't show in the kill message, only Wordbanked item names.

However, despite working when tested on Live, some sword names still show in global kill messages. Guessing there's a different NBT data structure causing the issue.

RedDevel2:
Can confirm it's inconsistent with how it sometimes does and sometimes doesn't use the item name.

okx-code:
It's only broken on admin-spawned items that contain colours in their name, it works fine for anvil renamed items

RedDevel2:
No, It's also for renamed ones. It'll show up dark gray same as rest of the text.

Diet-Cola:
Uhhh is this worth holding off fixing until 1.20.5 stuff in the dev chat?

PR #872, opened 2025-10-15
Custom items rework

### LINKED ISSUE #737: AddCustomModelData tags to All custom civ items.
Item such as Meteoric Iron and its tools/armor should be given custommodeldata tags so players can make resource packs that make these look different.

Huskydog9988:
Reworks the custom item interface as well as adds support for custom textures. Additionally, it migrates more custom items to proper custom items, for better fm config usability, plus removing the need to match items purely based on lore. 

Fixes #737


plugins/factorymod-paper/src/main/resources/plugin.yml
```diff
@@ -3,7 +3,7 @@ main: com.github.igotyou.FactoryMod.FactoryMod
 author: Maxopoly, igotyou
 version: ${version}
 depend: [ CivModCore ]
-softdepend: [ CivMenu, NameLayer, Citadel, Heliodor ]
+softdepend: [ CivMenu, NameLayer, Citadel, Heliodor, EssenceGlue, Bastion, HiddenOre ]
```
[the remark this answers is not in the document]
Diet-Cola: Agree, I guess the other alternative would be maybe a CustomItems plugins that plugins can depend on but depends on thoughts @okx-code

plugins/bastion-paper/src/main/java/isaac/bastion/BastionItems.java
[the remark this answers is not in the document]
Huskydog9988: I think so, but it would only get us so far. For very basic items like bastions and essence, which purely are just 'cosmetic' custom items, yea. For stuff like heliodor, I have no idea how you could even apply the block rules for example from a config without it getting messy fast. IMO, having it in the code is better, especially because these are supposed to be registered before **any** config loading is done since they can technically be used anywhere in the config.

plugins/factorymod-paper/src/main/java/com/github/igotyou/FactoryMod/FactoryMod.java
```diff
@@ -31,7 +32,9 @@ public void onEnable() {
 
     @Override
     public void onDisable() {
-        manager.shutDown();
+        if (manager != null) {
+            manager.saveFactories();
```
Diet-Cola: Did you remove the shutdown method?

Diet-Cola (approved):
Overall solid, downside is losing config driven custom items however this allows us to go further down the line. More worried about dependency stuff down the line if we stay with this approach however as mentioned in the comment

plugins/bastion-paper/src/main/java/isaac/bastion/BastionBlock.java
```diff
@@ -95,13 +96,7 @@ public double getErosionFromElytra() {
...
     public boolean inField(Location loc) {
         return !(yLevelCheck(loc) || // don't add parens if you don't know why it wasn't there.
-            (type.isSquare() &&
-                (Math.abs(loc.getBlockX() - location.getBlockX()) > type.getEffectRadius() ||
-                    Math.abs(loc.getBlockZ() - location.getBlockZ()) > type.getEffectRadius())) ||
-            (!type.isSquare() &&
-                ((loc.getBlockX() - location.getX()) * (float) (loc.getBlockX() - location.getX()) +
-                    (loc.getBlockZ() - location.getZ()) * (float) (loc.getBlockZ() - location.getZ()) >= type.getRadiusSquared()))
-        );
+            (type.isSquare() && (Math.abs(loc.getBlockX() - location.getBlockX()) > type.getEffectRadius() || Math.abs(loc.getBlockZ() - location.getBlockZ()) > type.getEffectRadius())) || (!type.isSquare() && ((loc.getBlockX() - location.getX()) * (float) (loc.getBlockX() - location.getX()) + (loc.getBlockZ() - location.getZ()) * (float) (loc.getBlockZ() - location.getZ()) >= type.getRadiusSquared())));
```
Metriximor: why remove the formatting that made this readable?

Huskydog9988: Yea your right, auto formatting error. I'll try and force the formatter to ignore.

Protonull:
It might be worth adding an item-tagging system while you're at it which would add an easy means to tag an item as compacted or as wordbanked. It could be as easy as storing a set of strings, that way testing for wordbank is just checking for the presence of `"wordbank"`, instead of testing the internal structure of the text component, which is so cursed and apparently faulty (#278).

PR #885, opened 2025-10-23
Give primary owners guaranteed access to solely-owner permissions

### LINKED ISSUE #148: Disabling 'PERMS' permission for Owner rank stops Primary Owner from editing group permissions
Disabling 'PERMS' permission for Owner rank stops everyone including the Primary Owner from editing group permissions which means the group's permissions get stuck as they are.

Protonull:
This should fix #148


Diet-Cola:
Not an issue but imo we should move "primary owner" to more being the group founder and phase out its weirdness with stuff like this

PR #892, opened 2025-11-15
Fix error when enchants.yml missing

MrJeremyFisher:
Defunct file. This will just error when the file isn't present.


Protonull:
You forgot to remove the file from the Ansible templates, see #792

PR #893, opened 2026-01-05
Rename locking

MrJeremyFisher:
Allows locking item names by combining them with honeycomb in an anvil. Locked items cannot be renamed by a player and cannot be wordbanked (though this is only necessary if items are locked with their default names as named items cannot be wordbanked). See my comments for bits I couldn't figure out but that would be nice, like changing the "Too Expensive!" text.


MrJeremyFisher:
Fixed combining items for repair with name-locked items and adding enchantments from books. Also fixed using >1 honeycomb casuing the rest of the honeycomb left in the anvil to be deleted.

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/NameLocker.java
```diff
@@ -0,0 +1,121 @@
...
+import org.bukkit.event.inventory.InventoryType;
+import org.bukkit.event.inventory.PrepareAnvilEvent;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.ItemMeta;
+import org.bukkit.persistence.PersistentDataContainer;
+import org.bukkit.persistence.PersistentDataType;
+
+public class NameLocker extends BasicHack {
+
+    final NamespacedKey lockedKey = new NamespacedKey(plugin, "name-locked");
```
okx-code: package private?

okx-code (changes requested):

PR #898, opened 2026-02-16
Name on hover

CitrineArbiter:
Please ignore all except the 3rd commit, im dumb and don't know how to remove the others.

Feature adds
1) a file to dump name history into from nlcpn
2) a shared volume to put that name history into
3) adds a thing to starman to display name changes in order on chat hover
4) adds a nlrpn command to remove a name from a player's name history using their current username
5) adds a whois command for any player to look up someones names based on their current one

I'm assuming the way its storing the name history isn't best practice but I believe best practice would probably be to dump it into the private config and idk how to do that


okx-code (changes requested):
Need to be moved to SQL based as discussed. Also try not to include changes to the docker compose file please

okx-code:
Feel free to re open if the above is addressed, but closing this for now

PR #899, opened 2026-02-18
Increase terracotta rates

azuziii:
Increased terracotta rates since factories should produce more than vanilla recipes


okx-code:
The productivity from the factory is the reduction of dye, not creating more terracotta by simply dying it

PR #906, opened 2026-04-06
Truncate Factory Recipes in GUI

Huskydog9988:
Added Diet-Cola's patch to truncate fm's gui so it doesnt go off the screen.
image


okx-code (changes requested):
Can you reduce the duplication here?

PR #907, opened 2026-04-06
add clickable dest signs

### LINKED ISSUE #690: Suggestion: Clickable /dest signs
by clicking on the sign, the destination command on the sign is automatically applied

Huskydog9988:
Dest signs are denoted with [DEST] or [DESTINATION] on the first line. A dest sign can have one dest per remain line on the sign. Players can punch the sign to cycle through the available dests. Destinations prefixed with "/dest " will have the prefixed removed automatically.

fixes: #690


Protonull:
imo, it should be `[DESTINATIONS]` to distinguish it from `[dest]`/`[!dest]` junction signs

okx-code (changes requested):
I agree this needs to be clearly distinguished from regular destination signs, like the first line should be `[set destination]` or something

Huskydog9988:
I wanted players to be able to use existing dest signs as they already exist, plus I don't feel the ambiguity is all that detrimental. I also think that just having a single "railswitch" sign is easier to understand then having multiple different kinds that end up just looking the same.

Protonull:
For what it's worth, I went with `[set destination]` for ClickDest *which definitely no one uses since it's illegal.* I could've gone with `[dest]` as you've done, it would give a layer of plausible deniability, but I do think gives a confusing dual purpose to those signs. I cannot read your mind, but I assume it's the same reason as why `[!dest]` signs aren't also supported in this PR despite having the exact same purpose (just a different redstone output).

okx-code:
Yeah Ori is right that the [!dest] signs don't quite fit into this as written, I would prefer [set destination]

Huskydog9988:
I think [set destination] is moving in the wrong direction. The more obtuse the directive is the less appeal I think it would have for players. If [dest] is such a problem, I'd rather [dests] / [destinations].

leadattic:
I think this could be distinguished in some way using colored signs. Maybe make clickable signs green and have the text be [applydest] or something.

PR #910, opened 2026-04-14
Fix: typos in bio_factory configuration

azuziii:


okx-code:
Why remove the concrete mixer kit recipe? In general I don't want to change stuff that works fine. You're also changing the music disc names by using a different escaping mechanism?

azuziii:
I thought you told me to remove the concrete mixer repair kits a few months ago

PR #912, opened 2026-04-14
move Migrator to separate library

Huskydog9988:
Moves migrator class to separate library. (Part of banstack rework preparation)


okx-code:
Please could you fold this PR into the banstick migration instead as it's a bit hard to see where this class will end up otherwise

PR #914, opened 2026-04-15
Fix typo in bio_factory configs

azuziii:


okx-code:
This doesn't really make any difference in how the factory works so closing this

PR #916, opened 2026-04-19
Fix emeralds blocking chest interactions

grepsedawk:
Holding an emerald previously cancelled right-click-block events on
any interactable block, preventing chests from opening at all. Mirror
vanilla food behavior instead: if the player is not sneaking, let the
block interaction proceed normally; if sneaking, consume the emerald
for XP as before.


xFier (approved):

ProgrammerGodMC (approved):

okx-code (changes requested):
This is a little better than cancelling for some reason (???), but I don't think it's quite the right fix. isInteractable is deprecated and I copied the logic from the NMS code into Citadel for a [similar fix](https://github.com/CivMC/Civ/blob/eb1c8623d31270fa5501bf2c0264f9e6a6c2263c/plugins/citadel-paper/src/main/java/vg/civcraft/mc/citadel/listener/BlockListener.java#L320-L325), which could likely be applied here instead.

PR #918, opened 2026-04-19
Repair the Gradle test suite and run it on PRs

grepsedawk:
The test suite was effectively dead. useJUnitPlatform() was never
configured in the parent plugins/build.gradle.kts, so every plugin's
test task discovered zero tests and reported BUILD SUCCESSFUL even
when src/test/ contained real JUnit 5 tests. The pinned JUnit (5.8.2)
also fell out of alignment with the platform launcher Gradle ships,
which would have failed discovery once useJUnitPlatform() was added
anyway. And civmodcore-paper's existing ItemMetaTests crashed during
class load because its static ItemStack(Material.STICK) field needed
a Bukkit Material registry that no test ever bootstrapped.

This change repairs the suite end to end:

- Enable useJUnitPlatform() once at the parent so every plugin
  inherits it. Bump JUnit to 6.0.3 with the platform launcher pinned
  to the same version ref so the engine and launcher stay aligned.
- Add MockBukkit to civmodcore-paper and bootstrap a ServerMock per
  test. Work around a paperweight quirk: the mojang-mapped server
  jar ships its own ServiceLoader providers for RegistryAccess,
  ServerBuildInfo and other Paper APIs, and Paper's loadAll() rejects
  duplicates. A small build-time task strips those entries from a
  copy of the mapped jar and substitutes it onto the test classpath,
  leaving NMS classes (which NBTTests needs) intact.
- Disable testBaseComponent — its assertion presumed a distinction
  between legacy setDisplayName(String) and modern displayName(
  Component) that no longer holds in current Paper.

Wire the result into CI: the existing Check All workflow now publishes
a JUnit report and uploads the HTML test results on failure, so PR
authors can see exactly what broke without re-running the build
locally.

The README gains a short Tests section pointing future contributors
at where tests live, how to run them, and how the MockBukkit /
paperweight workaround is set up. civmodcore-paper is now a working
template anyone can copy when adding tests to other plugins.


plugins/civmodcore-paper/build.gradle.kts
```diff
@@ -21,4 +21,47 @@ dependencies {
...
+// duplicates, so shadowing isn't enough — strip the entries from a copy of
+// the mapped jar and substitute that on the test classpath.
+//
+// The wildcards intentionally over-exclude under io.papermc.paper.** and
+// net.kyori.adventure.** to avoid having to update this list every time
+// paperweight ships a new bridge interface; MockBukkit re-registers the
+// providers it cares about, and anything else is unused at test time.
+//
+// TODO(paperweight>2.0.0-beta.17): no public accessor for the mapped server
+// jar exists yet — replace this internal-API reach-in if one is added.
```
okx-code: What is this? This doesn't make any sense, please don't submit slop

grepsedawk: Trimmed in the latest commit to a two-line note covering the one load-bearing fact: Paper's `loadAll()` rejects duplicate ServiceLoader providers, so when both paperweight's mapped server jar and MockBukkit are on the test classpath the build dies with `Expected to find one service interface ..., found multiple`. Stripping the Paper-side providers from the test-only copy of the jar is what lets MockBukkit's mocks win.

That whole task only exists to make MockBukkit usable here. Happy to drop the MockBukkit setup entirely if you'd rather not pull it in — civmodcore's existing `ItemMetaTests` would go back to being skipped (they were already not running before this PR because `useJUnitPlatform()` wasn't wired anywhere, which is the actually-load-bearing change). Let me know which way you want to go.

okx-code: I'd appreciate if you would talk to me -- it looks like this is a fundamental [incompatibility with mockbukkit](https://docs.mockbukkit.org/docs/en/user_guide/advanced/paperweight) and it's unlikely this workaround would actually fix anything

okx-code: It wasn't? That link literally describes how to fix it. Greps, you're better than this

okx-code: Sure, and you wouldn't happen to have a recipe for spaghetti bolognese as well? Need to make some dinner tonight. Anyway, let me know on Discord if you want to talk

adeadhead:
Why was this closer without resolution?

PR #920, opened 2026-04-22
fix dest setting not being saved

Huskydog9988:
The dest setting fails to save because the reset setting was sometimes overwriting it during the config save. This could have been prevented by not allowing duplicate settings keys from being registered.


okx-code:
bruh this was it?

okx-code:
Can we maybe prevent this from getting saved at all?

Huskydog9988:
All it adds is a single line to the config file because the value will always be empty for players. So imo its not worth reworking the settings api for something quite trivial.

okx-code (approved):

PR #926, opened 2026-05-25
Add support for waxed varients for CopperRail simple admin hack

bulbm1n:
https://discord.com/channels/912074050086502470/1508545613782843642

Per my own suggestion, pretty simple implementation that treats it as a regular copper block then turns it back to a waxed copper block after it runs the checks and needs to replace the block. Felt that was the easiest way.

### DIFF
--- a/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CopperRail.java
+++ b/plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CopperRail.java
@@ -41,9 +41,9 @@ public class CopperRail extends BasicHack {
     }
 
     private enum CopperStage {
-        UNAFFECTED(Material.COPPER_BLOCK, Material.WAXED_COPPER_BLOCK, 1.0f),
+        UNAFFECTED(Material.COPPER_BLOCK, Material.WAXED_COPPER_BLOCK, .75f),
         EXPOSED(Material.EXPOSED_COPPER, Material.WAXED_EXPOSED_COPPER, 1.0f),
-        WEATHERED(Material.WEATHERED_COPPER, Material.WAXED_WEATHERED_COPPER, 0.75f),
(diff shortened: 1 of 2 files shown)


okx-code (changes requested):
If you're going to encode the oxidation steps like this you might as well remove all NMS usage related to the oxidation. Code can be cleaned up significantly.

bulbm1n:
I left it using the current system so I didn't have to re-implement WeatheringCopper.getChanceModifier(). I can rewrite it though. I only read your original response from the notif, not your edited one, much easier fix.

bulbm1n:
Let me know how that is, I rewrote a lot of it. More self-contained

okx-code (changes requested):
You changed the axe damaging semantics so that it is no longer damaged for each oxidation step. also not sure why you removed the comment for the blockformevent listener?

plugins/simpleadminhacks-paper/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/CopperRail.java
```diff
@@ -47,6 +40,57 @@ public CopperRail(SimpleAdminHacks plugin, BasicHackConfig config) {
         super(plugin, config);
     }
 
+    private enum CopperStage {
+        UNAFFECTED(Material.COPPER_BLOCK, Material.WAXED_COPPER_BLOCK, 1.0f),
+        EXPOSED(Material.EXPOSED_COPPER, Material.WAXED_EXPOSED_COPPER, 1.0f),
+        WEATHERED(Material.WEATHERED_COPPER, Material.WAXED_WEATHERED_COPPER, 0.75f),
```
okx-code: also this is the wrong way round, it's only unaffected blocks that have a 75% chance

okx-code:
Looks better overall though

bulbm1n:
Fixed all of the above. Got confused with the comment and the durability because I repeatedly had to pick this up and put it down

bulbm1n:
@okx-code

PR #953, opened 2026-07-19
Stop coral from drying when reinforced

polysemic-mc:
Adds a check for reinforcement when coral tries to dry (fade). If the block is reinforced, don't let it turn to its dead variant.

For non-full blocks (e.g. fans, normal coral, wall fans) this checks the block it is on. I'm not sure if having this behavior is desirable for those blocks at all. I'll adjust it to only work on the full blocks if that'd be considered more fitting.

Discord thread: https://discord.com/channels/912074050086502470/1525198619509129347

### DIFF
--- a/ansible/files/paper-config/ops.json
+++ b/ansible/files/paper-config/ops.json
@@ -22,11 +22,5 @@
     "name": "zenron_",
     "level": 4,
     "bypassesPlayerLimit": true
-  },
-  {
-    "uuid": "1cff6d38-b01b-4d0c-b3c6-7a536f616920",
-    "name": "Polysemic",
-    "level": 4,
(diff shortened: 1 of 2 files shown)


okx-code:
Can we not use MaterialTags.CORAL, CORAL_FANS etc for this?

polysemic-mc:
I've updated it to use tags. Coral fans and plants seem to be included in `CORALS`, except for wall fans for some reason. And blocks are in a separate one. But there's no tag for dead corals, so this way of implementing it doesn't let me check the resulting block. I don't really expect Mojang to ever do anything silly with that, but feels a little dirty.

polysemic-mc:
(Don't mind the history rewrite, fucked up some things)

PR #954, opened 2026-07-20
Add shelves to carpentry factory

polysemic-mc:
Adds shelves to the carpentry factory. Uses the same proportion (24 logs -> 128 shelves) as are used for hanging signs, minus the iron, since the recipes have the same proportion in crafting tables. Can of course be adjusted.

Normal proportion is 6 stripped logs -> 6 shelves.

Discord thread: https://discord.com/channels/912074050086502470/1504946498956427314

### DIFF
--- a/ansible/files/paper-config/plugins/FactoryMod/config.yml
+++ b/ansible/files/paper-config/plugins/FactoryMod/config.yml
@@ -1316,66 +1316,77 @@ factories:
       - make_fence_gates
       - make_signs
       - make_hanging_signs
+      - make_shelves
       - make_trapdoors
       - make_wood_doors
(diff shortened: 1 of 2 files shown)


dxminic:
@polysemic-mc Add the copper variants to their factory too while you're at it 😇

polysemic-mc:
> @polysemic-mc Add the copper variants to their factory too while you're at it 😇

no because those don't exist
image

dxminic:
> > @polysemic-mc Add the copper variants to their factory too while you're at it 😇
> 
> 
> 
> no because those don't exist
> 
> image
> 
> 
> 
> 

made u look 🤪

PR #955, opened 2026-07-21
Add bamboo to carpentry factory

polysemic-mc:
Adds bamboo to carpentry factory, following same recipes as other woods.

Discord discussion: https://discord.com/channels/912074050086502470/1528695163348582540

Also adds shelves, can be removed if #954 does not get merged. Are adjusted if that gets adjusted.

As mentioned on discord, I don't really see a reason to add mosaic since they're trivial to craft and already have a 1:1 conversion in vanilla.

### DIFF
--- a/ansible/files/paper-config/plugins/FactoryMod/config.yml
+++ b/ansible/files/paper-config/plugins/FactoryMod/config.yml
@@ -1336,6 +1336,13 @@ factories:
       - make_jungle_hanging_signs
       - make_jungle_trapdoors
       - make_jungle_doors
+      - make_bamboo_fences
+      - make_bamboo_fence_gates
+      - make_bamboo_signs
+      - make_bamboo_hanging_signs
+      - make_bamboo_trapdoors
+      - make_bamboo_doors
+      - make_bamboo_shelves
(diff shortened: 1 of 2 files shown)


PillagingPagans:
"... already have a 1:1 conversion in vanilla."

In Vanilla, Bamboo Blocks are worth only two planks, as opposed to other types of wood that are worth 4. The factory outputs should also be halved compared to other wood types, otherwise you are giving a huge, relative buff, to Bamboo Blocks compared to other woods, based on the vanilla crafting rates.

polysemic-mc:
> In Vanilla, Bamboo Blocks are worth only two planks, as opposed to other types of wood that are worth 4. The factory outputs should also be halved compared to other wood types, otherwise you are giving a huge, relative buff, to Bamboo Blocks compared to other woods, based on the vanilla crafting rates.

Good call, updated recipes that would normally take planks to take 2x the amount of blocks. Not sure the same would be fitting for the ones that require full blocks (hanging signs, require iron anyway, and shelves)

The 1:1 conversion was a comment regarding the making of mosaic blocks.

PR #959, opened 2026-08-01
Add owner-controlled arena gravity

uxvi:
Let arena owners switch between Main and Zorweth gravity before another player joins. Apply the selected gravity to survival players and restore Main gravity when they leave the arena or stop playing.

Show the setting on each arena entry and let its owner toggle it with a right click. Keep ranked arenas and spectators on Main gravity, and cover the gravity values, parsing, toggling, defaults, and lock behavior.


catluver21321312-create:
This is awesome!

sewingtailnet:
very cool

okx-code:
Does not follow LLM disclosure policy. Arena locking should be removed as well as unit tests

okx-code:
Loaded arena should stay immutable with arenamanager reconciling state rather than the command directly updating the gravity via the listener

okx-code (changes requested):
see above

PR #965, opened 2026-09-01
Custom packets for ItemExchange trade displays

Protonull:
This allows for mods like TradeX to more accurately track shop chests and their offerings.


okx-code:
Hey gson is thread safe (mentioned in the navadocs) so you don't need to synchronise on it

okx-code:
Also with the namespacedkey I feel like it's a lot of complexity to handle this memoisation, is there a way to reduce it? Perhaps hardcoding the plugin name or dependency injecting it.

Protonull:
> Also with the namespacedkey I feel like it's a lot of complexity to handle this memoisation, is there a way to reduce it?

It's probably just class-loading neurosis on my part, but I fret over these kinds of static fields which are set upon class initialisation, which could potentially happen prior to `ItemExchangePlugin`'s constructor being called and its instance field set. I'm probably fretting over nothing.

PR #966, opened 2026-09-04
Add pale oak saplings to SaplingHack

polysemic-mc:
It's currently already in the config to drop more saplings, but it doesn't work because the leaves hadn't been mapped to the saplings in code yet.


RedDevel2:
oops

